PR #2, opened 2021-07-24
Add macro comment for MixScope

sbis04:
I have added the **macro comment tag** to the constructor of `MixScope` so that the documentation shows up in the IDE.

Have also updated the **Example** under that to the following:

```dart
MixScope<MixStyles>(
  data: MixStyles(),
  child: MixDescendantWidget(),
)
```

Is `MixDescendantWidget()` alright as the child or should I change it to `MyApp()`. Let me know how you want to keep the example.

### DIFF
--- a/lib/src/helpers/mix_scope.dart
+++ b/lib/src/helpers/mix_scope.dart
@@ -79,6 +79,8 @@ abstract class Model extends Listenable {
   }
 }
 
+/// {@template mix.mixscope}
+///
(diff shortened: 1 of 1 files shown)


leoafarias:
@sbis04 I believe `MyApp()` would be better as an example.

leoafarias (approved):

leoafarias:
I am closing this PR, as I have removed the Scope for now, as it doesn't seem that is needed for a basic implementation.

Issue #1, opened 2021-07-06
Widget: Pressable Component


leoafarias:
There is a need to create components that can be interacted with but easily styled. Right now this is not a problem as you can style all the components and wrap them in an InkWell or a GestureDetector, however, the creation of a `Pressable` component opens up some other possibilities.
- Ability to provide styling for interactions, `tap`, `hover`, `disabled` etc.
- Clean syntax for creating buttons and other pressable components using Mix

bdlukaa:
This should be something like:

```dart
PressableMix(
  Mix(
    disabled(* DISABLED MIX HERE *),
    pressing(* PRESSING MIX HERE *),
    hovering(* HOVERING MIX HERE *),
  )
)
```

leoafarias:
@bdlukaa that is a great suggestion for the API. Right now the Mix Widgets are stateless, but I am thinking this one will have to be stateful in order to implement those states.

bdlukaa:
Indeed. I already have something, I can make a PR

leoafarias:
#3 merged

PR #3, opened 2021-09-08
Pressable Widget

bdlukaa:
#1

Added `DisabledAttribute`, `FocusedAttribute`, `HoveringAttribute`, `PressingAttribute`.

```dart
Mix(
    apply(_container),
    pressing(Mix(bgColor(RxColors.primary))),
    hovering(Mix(bgColor(RxColors.secondary))),
    apply(RxShadows.shadow20),
)
```

I had to make `PressableMixerWidget` a stateful widget because the current api doesn't offer any type of state management.

### DIFF
--- a/lib/src/widgets/primitives/pressable.dart
+++ b/lib/src/widgets/primitives/pressable.dart
@@ -5,6 +5,15 @@ import '../../mixer/mixer.dart';
 import '../mix_widget.dart';
 import 'box.dart';
 
+enum _PressableState {
+
+  disabled,
+  focused,
+  hovering,
+  pressing,
+
+}
+
 class Pressable extends MixWidget {
   const Pressable(
     Mix mix, {
--- a/lib/src/widgets/primitives/pressable.dart
+++ b/lib/src/widgets/primitives/pressable.dart
@@ -147,8 +147,6 @@ class _PressableMixerWidgetState extends State<PressableMixerWidget> {
                 if (_shouldShowFocus && focused != null) return focused.mix;
               }();
 
-              print(hovering);
-
               return BoxMixerWidget(
                 widget.mixer,
                 child: () {
--- a/lib/src/widgets/primitives/pressable.dart
+++ b/lib/src/widgets/primitives/pressable.dart
@@ -5,15 +5,6 @@ import '../../mixer/mixer.dart';
 import '../mix_widget.dart';
 import 'box.dart';
 
-enum _PressableState {
-
-  disabled,
-  focused,
-  hovering,
-  pressing,
-
-}
-
 class Pressable extends MixWidget {
   const Pressable(
     Mix mix, {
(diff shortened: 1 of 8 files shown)


lib/src/widgets/primitives/pressable.dart
```diff
@@ -58,12 +60,108 @@ class PressableMixerWidget extends MixerWidget {
...
+              return BoxMixerWidget(
+                widget.mixer,
+                child: () {
+                  if (mix != null) {
+                    return BoxMixerWidget(
+                      Mixer.build(context, mix),
+                      child: widget.child,
+                    );
+                  }
+                  return widget.child;
```
bdlukaa, line 159: This, for some reason, is not being recognized. Means the box is not being painted if `pressed(Mix(...))` is used, for example.

Any clues?

leoafarias: @bdlukaa Not sure exactly but looking at the code here are a few things. 
- I believe the MixWidget will have to be the one that is stateful.
- `Mixer.build()` allows to build the Mix based on the context, once the `mixer` is created I can be passed to a stateless widget and not modified.

Which means I think the logic need to be the following

```dart
/// In mix widget
Mix? gestureMix = () {
    if (!enabled && disabled != null) return disabled.mix;
    if (_pressing && pressing != null) return pressing.mix;
    if (_hovering && hovering != null) return hovering.mix;
    if (_shouldShowFocus && focused != null) return focused.mix;
}();

final mixer = Mixer.build(context, Mix.combine(mix, gestureMix));

```

Before running the `Mixer.build()` you will have to combine the gesture mix with the "original" mix passed into the pressable. The reason why is that this Mix, might contain styling information that you want to inherit, and use the gesture to `override`. 

Also, I believe there should be only 1 BoxMixer widget, and there is no need for the conditional check using the workflow I described above.

leoafarias: Also one more thing, you are massing the `mix` into the attribute, we might want to consider passing an Attribute itself, similar to what is done on the `DarkModeAttribute`, you can find it here `dark_mode.dart`. 

This would allow passing an attribute, however, you can pass a mix by using the `apply` attribute, so the API could look something like this.

```dart
hovering(bgColor(Colors.green);
```
But would also allow for more robust implementations like this

```dart
hovering(apply(disabledMix, otherMix));
```

bdlukaa: I don't think that's a good idea. We can't use two `hovering` parameters because one will cancel the other

leoafarias: Mixes can be composable, so they won't cancel out each other but will override properties. For example, a button can have an overall shape (rounded), margin and padding, together with a shadow, maybe you only want to override the shadow, or you want the color, but you don't want to compose all the basic shape again, this allows to easily override based on the order you call

PR #7, opened 2021-09-17
Animation

### LINKED ISSUE #4: Feature: Animations
Provide apis for animations. Something like:

```dart
AnimatedMix(Mix(
  bgColor(Colors.green),
))
```

Whenever the Mix is changed, it should be animated.

bdlukaa:
#4 

---

This PR adds animations to attributes. The API is simple:

```dart
Mix(
  bgColor(Colors.green).animated(),
),
```

On a `Pressable` widget:

```dart
Mix(
    apply(_container),
    bgColor(Colors.primary).animated(),
    disabled(Mix(bgColor(Colors.grey.shade300).animated())),
    hovering(Mix(bgColor(Colors.primary.shade400).animated())),
    pressing(Mix(bgColor(Colors.primary.shade600).animated())),
    apply(RxShadows.shadow20),
);
```

The background color will change according to the mouse state.

---

An example of how to add animation to a color attribute:

```dart
class BackgroundColorAttribute extends Attribute<Color> with AnimatedMix<Color> {
  BackgroundColorAttribute(this.color);

  final Color color;

  @override
  Color get value => color;
}
```

Just add a `with AnimatedMix<Color>` to the class.

### DIFF
--- a/lib/src/mixer/mixer.dart
+++ b/lib/src/mixer/mixer.dart
@@ -389,21 +389,21 @@ class Mixer {
 
     // This handles in case there is a color attribute
     // but not other decoration
-    if (mixer.border != null ||
-        mixer.borderRadius != null ||
-        mixer.boxShadow != null) {
-      mixer.decoration = BoxDecorationAttribute(
-        BoxDecoration(
-          border: mixer.border?.value,
-          borderRadius: mixer.borderRadius?.value,
-          color: mixer.backgroundColor?.value,
-          boxShadow: mixer.boxShadow != null ? [mixer.boxShadow!.value] : null,
-        ),
-      );
+    // if (mixer.border != null ||
+    //     mixer.borderRadius != null ||
+    //     mixer.boxShadow != null) {
+    //   mixer.decoration = BoxDecorationAttribute(
+    //     BoxDecoration(
+    //       border: mixer.border?.value,
+    //       borderRadius: mixer.borderRadius?.value,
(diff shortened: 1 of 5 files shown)


lib/src/mixer/mixer.dart
```diff
@@ -401,6 +401,10 @@ class Mixer {
         ),
       );
 
+      if (mixer.backgroundColor?.hasAnimation ?? false) {
+        mixer.decoration!.animated();
+      }
```
bdlukaa, line 406: This is not the best approach IMO, but it seems it's the only option for now. Animating the whole decoration because only one property (color) is animated is not a good thing.

lib/src/widgets/primitives/box.dart
```diff
@@ -82,17 +82,35 @@ class BoxMixerWidget extends MixerWidget {
...
+          curve: mixer.backgroundColor!.animationCurve!,
+          color: mixer.backgroundColor!.value,
+          child: current,
+        );
+      } else {
+        current = ColoredBox(
+          color: mixer.backgroundColor!.value,
+          child: current,
+        );
+      }
```
bdlukaa, line 97: I didn't find a way to animate solely the `ColoredBox`. We might want to create a widget that animates animated mixes (`AnimatedColoredBox` and `AnimatedDecoratedBox`, for example).

bdlukaa (commented):
In a next PR I'll implement the animations for the appliable mixes.

PR #11, opened 2021-09-27
Fix IconMix

bdlukaa:
Currently IconMix throws an error is `iconSize` attribute is not specified.

This also fix the issues on the tests caused by the animation api!


leoafarias (approved):
LGTM

PR #28, opened 2021-12-26
Initial doc changes

rickbsgu:
I've worked over primarily the 'Concepts' page, for starters.  Also changed some of the header styling to something I think works better.

Might have gotten carried away.

Look it over and we can discuss.

Do you have a git Project set up for this?

thanx,
rickb

### DIFF
--- a/website/pages/_document.js
+++ b/website/pages/_document.js
@@ -17,6 +17,10 @@ class MyDocument extends Document {
             href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;700;800&display=block"
             rel="stylesheet"
           />
+          <link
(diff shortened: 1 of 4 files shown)


leoafarias:
@rickbsgu thank you for the changes I think they are great. I will merge and only remove some of the styling changes as it will be part of a bigger adjustment which we can discuss.

rickbsgu:
Hey, Leo -

Sounds good.

The styling changes were ad-hoc just to provoke some thought.  They would
have to be done better, especially the outdents.  I need to study nextjs to
understand how it handles responsiveness.

I'll merge back and tackle the other pages content.

My goal is to understand the package.  Figure helping out with the docs
would be a good way to do that.

Cheers,
rickb

PR #61, opened 2022-03-16
Implement BoolExtension

bdlukaa:
Makes the following syntax possible:

```dart
final mix = Mix(
  (true)( // [condition]
    bgColor(Colors.blue), // this is only executed if [condition] is true
  ),
);
```

### DIFF
--- a/demo/lib/app_shell.dart
+++ b/demo/lib/app_shell.dart
@@ -43,36 +43,36 @@ class AppShell extends HookConsumerWidget {
         ),
         body: Row(
           children: [
-            // NavigationRail(
-            //   extended: true,
-            //   selectedIndex: selected.value,
-            //   onDestinationSelected: (index) {
-            //     selected.value = index;
-            //   },
-            //   // labelType: NavigationRailLabelType.selected,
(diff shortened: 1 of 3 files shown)


leoafarias:
@bdlukaa we currently have the `Mix.chooser()` which you can do the following. 

```dart
final chooseSecondMix = Mix.chooser(
        condition: false,
        ifTrue: firstMix,
        ifFalse: secondMix,
      );
```

Do you want to share a bit more about the usecase?

bdlukaa:
`Mix.chooser` chooses between two `Mix`es. By using `BoolExtension`, a `Variant` can be used inside a `Mix` only if the condition is true. Makes the following possible:

```dart
class _MyState extends State<My> {

  bool _hasError = false;

  @override
    final mix = Mix(
      borderRadius(6.0),
      elevation(4.0),
      textStyle($h1),
      bgColor(Colors.grey), // Default color is grey
      (_hasError)(bgColor(Colors.red)), // But if it has error, color is red
    );
    
    return ...;
  }

}
```

leoafarias (approved):
LGTM

PR #71, opened 2022-03-22
"Variants rework" documentation

bdlukaa:
Documentation for #70

### DIFF
--- a/website/pages/docs/core-concepts/variants.mdx
+++ b/website/pages/docs/core-concepts/variants.mdx
@@ -137,6 +137,31 @@ MixTheme(
 - `dark()`: Applies attributes when the theme is in dark mode.
 - `light()`: Applies attributes when the theme is in light mode.
 
+### State Holding
+
+- `active()`: Applies attributes when the widget state is marked as active
+- `inactive()`: Applies attributes when the widget state is marked as inactive
+
+### Text Styling
+
+For widgets such as Dialogs, Cards and Modals, it's good to have titles and content.
+
+- `title()`: Applies attributes when the `TextMix` is marked as title
+- `paragraph()`: Applies attributes when the `TextMix` is marked as paragraph
+
+```dart
+final mix = Mix(
+  title(
+    titleCase(), // Makes the title uppercase
+    font(
+      weight: FontWeight.bold, // Makes the title bold
+    ),
+  ),
+);
+```
+
+All other texts that are not marked as `title` will not be touched.
+
 ## Custom Variants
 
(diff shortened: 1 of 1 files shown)


lib/src/attributes/variants/variants.utils.dart
```diff
@@ -132,4 +130,28 @@ class VariantUtils {
...
+    return Variant<T>(SystemVariants.inactive.value);
+  }
+
+  /// Short Utils: title
+  static Variant<T> title<T extends Attribute>() {
+    return Variant<T>(SystemVariants.title.value);
+  }
+
+  /// Short Utils: paragraph
+  static Variant<T> paragraph<T extends Attribute>() {
```
leoafarias, line 150: Don't think these variants should be  here. Maybe they should get defined within the headless component.

lib/src/attributes/variants/variants.utils.dart
```diff
@@ -132,4 +130,28 @@ class VariantUtils {
...
+    return Variant<T>(SystemVariants.active.value);
+  }
+
+  /// Short Utils: inactive
+  static Variant<T> inactive<T extends Attribute>() {
+    return Variant<T>(SystemVariants.inactive.value);
+  }
+
+  /// Short Utils: title
+  static Variant<T> title<T extends Attribute>() {
```
leoafarias, line 145: Same for this variant. I believe any variant for business logic implementation that is not generic for use in the primitives should not be here

lib/src/attributes/variants/variants_short.utils.dart
```diff
@@ -12,7 +12,16 @@ final landscape = VariantUtils.landscape();
 final dark = VariantUtils.dark();
 final light = VariantUtils.light();
 
+final enabled = not(disabled);
 final disabled = VariantUtils.disabled();
 final focus = VariantUtils.focused();
 final hover = VariantUtils.hover();
 final press = VariantUtils.pressing();
+
+final active = VariantUtils.active();
```
leoafarias, line 21: What is an active variant also. Primitives don't have the concept of active, maybe this is part of the headless conversation

leoafarias (changes requested):
I think we should hold a bit on the headless as it's getting it's way into the core of mix.

Let's have a conversation on how to separate that

bdlukaa:
@leoafarias do you think #70 should be reverted, then?

leoafarias:
@bdlukaa will review later and maybe just separate.. give me a couple of days

leoafarias:
@bdlukaa can we revert #70, and keep all the headless specific things separate?

PR #74, opened 2022-04-28
Revert "Variants rework"

bdlukaa:
Reverts fluttertools/mix#70
See https://github.com/fluttertools/mix/pull/71 for more info


lib/src/headless/radio_button.dart
```diff
@@ -37,6 +37,9 @@ class RadioButtonX extends RemixableWidget {
   final ValueChanged<bool>? onChanged;
   final RadioButtonIndicator indicator;
 
+  static Variant whenOn = const Variant('active');
```
leoafarias, line 40: Can we rename these to match the variant name?

leoafarias (changes requested):

PR #79, opened 2022-05-05
API Clean-up and reorganization

leoafarias:


bdlukaa:
I believe https://pub.dev/packages/window_manager is better. It is constantly updated!

PR #95, opened 2022-10-04
Turn Variant const again

tilucasoli:
When the new version of Mix was released, it came with break changes, one of them is the class `Variant` hasn't a constant constructor anymore, what don't allow to use the new format of enums with properties.

In this PR I fix this and `Variant` constructor become constant again.

Changing this:
```dart
enum SomaButtonAppearance {
  primary(Variant("primary")),
  secondary(Variant("secondary")),
  ghost(Variant("ghost"));

  const SomaButtonSize(this.variant);

  final Variant variant;
}
```

To
```dart
enum SomaButtonAppearance{
  primary,
  secondary,
  ghost;

  Variant get variant {
    switch (this) {
      case primary:
        return Variant("primary");
      case secondary:
        return Variant("secondary");
      case ghost:
        return Variant("ghost");
    }
  }
}
```

### DIFF
--- a/lib/src/attributes/variants/variants.utils.dart
+++ b/lib/src/attributes/variants/variants.utils.dart
@@ -136,6 +136,6 @@ class VariantUtils {
   }
 
   static Variant<T> not<T extends Attribute>(Variant<T> other) {
-    return other..inverse = true;
+    return other.inverseInstance();
   }
 }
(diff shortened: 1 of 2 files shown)


leoafarias (approved):
LGTM

leoafarias:
@all-contributors please add @tilucasoli for code

Issue #93, opened 2022-09-16
[proposal] Replace mix theming with material theme extensions


bdlukaa:
We can take advantage of Material's theme extensions to replace the current mix theming. Currently, we define global variables and attach value to them in a `MixThemeData`, so that we can use globally.

**Proposal**

A callback based function `theme<T extends ThemeExtension>((T))` that can be called inside a `Mix`. The mixer will look into every attribute for a `theme` and perform the operation based on the current context.

Something like:

```dart
final mix = Mix(
  bgColor(theme<MyColors>((myColors) => myColors.brandColor)),
  textColor(theme<MyColors>((myColors) => myColors.danger))),
);

// This is from the official example. Link attached below
@immutable
class MyColors extends ThemeExtension<MyColors> {
  const MyColors({
    required this.brandColor,
    required this.danger,
  });

  final Color? brandColor;
  final Color? danger;

  ...
}
```

**Other things I have considered**
Making `theme` more readable and usable with something like

```dart
final mix = Mix(
  bgColor(theme<MyColors>().brandColor),
  textColor(theme<MyColors>().danger),
);
```

But I can't seem to find a way where `theme` would have access to the current context

**Additional context**
https://api.flutter.dev/flutter/material/ThemeData/extensions.html

leoafarias:
@bdlukaa will review in detail. A few things to consider is that Mix should not be dependent on only Material design tokens, they also will need to be able to use them and/or create custom ones for their own design system.

leoafarias:
@bdlukaa I like this approach and could simplify a lot of our theming logic. I wonder if we can use this for spacing, sizes and so on. Could you do a POC for the Space theme data, and we can validate the implementation

leoafarias:
@bdlukaa while taking a closer look into this, I don't think ThemeExtensions replace the Mix Tokens, especially if you are trying to accomplish something like `theme<MyColors>().danger`, I am going to go through a bit and see if I can make any advance on the separation of Mix Tokens and the ThemeExtensions

leoafarias:
We have rewritten large parts of Mix, which better supports a lot of Material Theme features. However, I have renamed this to highlight better support for ThemeExtensions.

PR #99, opened 2022-10-07
[POC] Material theme extensions

bdlukaa:
#93

The class `MixProperty<T>` was created, which will be in every mix attribute. With it, it's possible to use values that depends on the current `BuildContext`. This means it's possible to do both `bgColor(Colors.white)` and `bgColor(theme<MyColors, Color>((colors) => colors.white))`.

Added `theme`, a function that fetches `ThemeExtension`s from the closest `Theme` widget. Also a basic example was added to the `/demo` app. To try it, change the app theme mode (light/dark) and see the "Gradient Box" color change as the theme changes.

Another advantage of this approach is that we now can make use of Material's `AnimatedTheme`. When a property in your `ThemeExtension` changes, its `lerp` function is called and is animated accordingly to the theme. _(`animate()` works with it too, in case you want to delay the animation)_

_For now, only the following utils are affected: `bgColor`, `margin` and `height`_


bdlukaa:
Even if we don't want to make use of `ThemeExtension`s in mix directly, I believe adding a way to access the `BuildContext` is a must.

However, with the implementation above, we lose the dart type-safety (by using `dynamic`). 

@leoafarias I'd love to hear your thoughts on this

leoafarias:
Thanks for the work on this @bdlukaa. We have already 2 abstractions for values of context. `Dto` and `Tokens` (Which might rename `Refs`, want to see how these fit together.

Also the example allows to pass properties directly to a Box, but since Box is a primitive, I don't think a very similar properties is likely on the primitives themselves.  For example on the `headless` widgets this might be different.

I believe this extensions should work similar to the tokens, or see if we can abstract the tokens to take advantage of `ThemeExtensions` instead of a separate values on MixTheme.

PR #100, opened 2022-10-09
Refactor & Clean-up

leoafarias:
- Refactored MixContext to allow for more dynamic Attribute conversion.
- Simplified props conversion
- Reorganized folder structure

## Breaking Changes
- InheritedAttribute is not a Generic anymore


lib/src/mixer/mix_context.dart
```diff
@@ -1,72 +1,55 @@
...
-  final FlexProps flexProps;
-  final ZBoxProps zBoxProps;
-
   /// Used to obtain a [InheritedAttribute] from [InheritedAttributes].
   ///
   /// Obtain with `mixContext.fromType<MyInheritedAttribute>()`.
-  T fromType<T extends InheritedAttribute<T>>() {
-    final attribute = inheritedAttributes.fromType<T>();
-    debugCheckInheritedAttribute<T>(attribute);
+  T? fromType<T extends InheritedAttribute>() {
```
pbissonho: @leoafarias What do you think about keeping the fromType with a non-nullable return and create a new method for nullable returns? Something like maybeFromType.

would be:

// Not return null and has an assert with a formatted error message.
// This working like other libs as Provider with Inherited widgets.
```dart
T fromType<T extends Inherited widgets >()
```
and

// Return null if there is not a mix for the specified type
```dart
T? maybeFromType<T extends InheritedAttribute>()
```

Thinking about the developer experience when creating and using new InheritedAttributes outside of the Mix. I believe that a check if it is null and an assert with an error message is important.

pbissonho (changes requested):

lib/src/mixer/mix_context.dart
leoafarias: @pbissonho We can create that, however should we just throw an error in case there are no attribute being returned from that type?

pbissonho: I understand that yes, it seems to be a good approach.

It would be something like:

```dart
T fromType<T extends InheritedAttribute<T>>() {
    final attribute = attributes.fromType<T>();
  
    if (attribute is! T) {
      throw '''
     No $T could be found starting from the mixContext 
     when call mixContext.fromType<$T>(). This can happen because you 
     have not create a Mix with $T.
    ''';
    }
  
    return attribute!;
  }
} 

```

## Example with non-nullable return:


```dart
 @override
 
    return MixContextBuilder(
      mix: myMix,
      builder: (context, mixContext) {
        //With non-nullable Mix handles the return and throw an error message
        //with a description if not have the attribute in the mixContext.
        final inputTheme = mixContext.fromType<InputDecorationThemeAttribute>();

        return TextFormField(
          decoration: InputDecoration(
            labelText: props.label,
            suffixIcon: const Icon(Icons.telegram),
            filled: true,
          ).applyDefaults(inputTheme),
        );
      },
    );
  }

```

## Example with nullable return:

```dart

    return MixContextBuilder(
      mix: myMix
      builder: (context, mixContext) {
        // With nullable I need to handle the return manually and I don't have an null reference error handle

        final inputTheme =
            mixContext.maybeFromType<InputDecorationThemeAttribute>()!;

        return  TextFormField(
              decoration: InputDecoration(
                labelText: props.label,
                suffixIcon: const Icon(Icons.telegram),
                filled: true,
              ).applyDefaults(inputTheme),
            ),
          );
      },
    );
  }
```

PR #104, opened 2022-11-08
`when` variant

### LINKED ISSUE #102: [Proposal] `when` variant
Mix already has a [long set](https://github.com/conceptadev/mix/blob/466cbce6316fd3970ca6aa3f0a2da646dd57591f/lib/src/variants/variants.dart#L7-L19) of built-in variants. Yet, there isn't a variant to check if a value is true or false: there's where the `when` variant would come in. It'd be similar to the `if` keyword in dart, taking a `bool` as a parameter.

**Basic usage**
```dart
// in a stateful widget

bool hasError = false;

void fetch() {
  try {
    // perform the fetch
    setState(() => hasError = false);
  } catch (e) {
    // handle the error
    setState(() => hasError = true);
  }
}

final mix = Mix(
  when(hasError)(
    bgColor(theme<MyColors>((colors) => colors.errorBackgroundColor)),
  )(
    // this would be the `else` case
    bgColor(theme<MyColors>((colors) => colors.surfaceBackgroundColor)),
  ),
);
```

**Why?**
Currently, it's very verbose to check if a value is true or false in a Mix. We must either ¹create a variant and use `mix.withVariant` or ²create inline conditional checks `hasError ? bgColor(...) : bgColor(...)`

bdlukaa:
Fixes #102

#### Basic usage:

```dart
// in a stateful widget

bool hasError = false;

void fetch() {
  try {
    // perform the fetch
    setState(() => hasError = false);
  } catch (e) {
    // handle the error
    setState(() => hasError = true);
  }
}

final mix = Mix(
  when(hasError)(
    rounded(5),
  )(
    // this would be the `else` case
    rounded(10),
  ),
);
```


leoafarias:
@bdlukaa Since Mix is made to be used outside of the `BuildContext`. It would be important to make this usable outside of it also or else it would be better to compose using Mix.chooser, or some other composable Mix API. 

A solution for this would be to accept a function that returns a boolean and passes context as a parameter. But I believe this then becomes a DynamicVariant. 

Maybe I am overthinking, but I am thinking of having a few ways of doing certain things on Mix, this could be a helper, but not sure if should be the main way to do conditionals within Mix.

bdlukaa:
@leoafarias 

> Since Mix is made to be used outside of the BuildContext. It would be important to make this usable outside of it

`when` should not provide a context, nor is required to be used inside a `BuildContext`. It depends only on a `bool` parameter.  For example, to warn the dev he's in debug mode, they could do the following:
```dart
final mix = Mix(
  when(kDebugMode)(
    bgColor(Colors.red),
  )(
    bgColor(Colors.blue),
  ),
);
```

It looks much better than
```dart
final mix = Mix(
  kDebugMode ? bgColor(Colors.red) : bgColor(Colors.blue),
);
```
It is also easier to maintain it throughout time

---

> A solution for this would be to accept a function that returns a boolean and passes context as a parameter

`Mix.chooser` has the downside of having to create two mixes. Imagine just wanting to change the background to red if an error happens. In this situation, the dev would have to do:
```dart
final errorMix = Mix(bgColor(Colors.red));
final successMix = Mix(bgColor(Colors.green));

final mix = Mix.chooser(
  condition: hasError,
  ifTrue: errorMix,
  ifFalse: succesMix,
);
```
_example from the [official docs](https://www.fluttermix.com/docs/concepts/mixing#mixchooser)_

On the other hand, when using the `when` variant, the code looks a lot more cleaner:
```dart
final mix = Mix(
  when(hasError)(
    bgColor(Colors.red),
  )(
    bgColor(Colors.green),
  ),
),
```

Another benefit from using `when` is that we can nest variant conditionals, making the mix composability ~(?)~ a lot better!

---

PR #109, opened 2022-12-21
Pressable and Gesture Variant optimizations

leoafarias:
- Optimizations on implementation on Gesture variants
- Changes on API to replace `Pressable`


Maus3rSR:
Hello !

I can still press the button, getting onPress function called, but without seeing the press style applied (it's really on a gentle press).
On TextButton from material I can see the visual press effect applied in the same time that the onPressed callback.

I don't know how to help you about this issue...

Some other things I noticed about api changes:
- applyMaybe
- disabled / press mix variant

I think these methods were not be exported because when I change to the new ones flutter cannot find them (on the other hand strangely my IDE can access the package code)

leoafarias:
@Maus3rSR will see if I add some delay; the reason for that is the tap effect happens very fast. The MaterialButton uses Inkwell, which has its own animation delay already implied. Will review, to see what can be done, maybe implement the InkWell might be an option.

leoafarias:
> * applyMaybe

The `applyMaybe` should have a deprecation notice, now the method is called `maybeApply`, also the `disabled` is called `onDisabled`, all Context variants have the `on` word as a best practice recommendation

PR #120, opened 2023-04-03
Attributes renamed from mix to style using the Mix alias StyleMix

marcobraghim:


lib/src/variants/variant_attribute.dart
```diff
@@ -9,14 +9,15 @@ import 'variant.dart';
 class VariantAttribute<T extends Variant> extends Attribute
     with MergeableMixin<VariantAttribute<T>>, EquatableMixin {
   const VariantAttribute(
-    this.variant,
-    Mix mix,
-  ) : _mix = mix;
```
leoafarias, line 14: This is changing the API signature from positional to named parameters. Since you area already changing the signature there is no need to deprecate it, you can just remove it, this is an internal class so it should be fine.

lib/src/variants/variant_attribute.dart
```diff
@@ -47,19 +48,21 @@ class VariantAttribute<T extends Variant> extends Attribute
...
-    ContextVariant variant,
-    Mix mix,
-  ) : super(
+    ContextVariant variant, {
+    @Deprecated('Use the style parameter instead') Mix? mix,
+    StyleMix? style,
+  }) : super(
           variant,
-          mix,
+          mix: mix,
```
leoafarias, line 64: Lets use `super.param` to pass down parameters to the super class

test/factory/mix_values_test.dart
```diff
@@ -28,10 +28,13 @@ void main() {
...
     final variantList = <VariantAttribute>[
-      VariantAttribute(const Variant('testVariant'), RandomGenerator.mix()),
+      VariantAttribute(
+        const Variant('testVariant'),
+        style: RandomGenerator.mix(),
+      ),
       VariantAttribute(
         const Variant('anotherTestVariant'),
-        RandomGenerator.mix(),
+        style: RandomGenerator.mix(),
```
leoafarias, line 37: Let's change the signature of `VariantAttribute` to a positional parameter, and since its an internal class, we should be fine.

leoafarias (changes requested):

leoafarias (approved):
LGTM

PR #138, opened 2023-07-11
Prepare for 1.0

aguilaair:
This PR removes some deprecated behavior (nested style processing) as well as fixes other deprecations.

Also increases Variants test coverage


lib/src/factory/style_mix_data.dart
```diff
@@ -19,16 +18,14 @@ class StyleMixData with EqualityMixin {
   });
 
   /// Creates a new [StyleMixData] instance from the provided [Iterable] of [StyleAttribute]s.
+  /// No longer expands nested attributes.
   factory StyleMixData.create(Iterable<StyleAttribute> attributes) {
-    //TODO: Remove expansion of nested attributes later
-    final expanded = _expandNestedAttributes(attributes);
```
leoafarias, line 24: Making a note of this to review that is ready to be removed

lib/src/dtos/border/border_side.dto.dart
```diff
@@ -67,8 +67,7 @@ class BorderSideDto extends Dto<BorderSide> {
       color: color?.resolve(mix) ?? _default.color,
       width: width ?? _default.width,
       style: style ?? _default.style,
-      // TODO: Temporarily removed because of compatibility issues
-      // strokeAlign: strokeAlign ?? _default.strokeAlign,
+      strokeAlign: strokeAlign ?? _default.strokeAlign,
```
leoafarias, line 70: Same here review compatibility impact

pubspec.yaml
```diff
@@ -5,7 +5,7 @@ homepage: https://github.com/leoafarias/mix
 
 environment:
   sdk: ">=2.17.0 <3.0.0"
-  flutter: ">=2.0.0"
+  flutter: ">=3.3.0"
```
leoafarias, line 8: We might want to review this constraint as it might impact adoption.

aguilaair: This constraint was added for the `strokeAlign` parameter, which hit stable on 3.3.0. It can always be dropped down, but users may encounter issues then

leoafarias (approved):
LGTM

PR #152, opened 2024-01-10
rewriting gap feature

### LINKED ISSUE #150: Improvement: Refactor Gap widget
The current version of the Gap widget has a lot of complexity in its behavior. A more straightforward implementation would be to use `Padding` instead so it would add margins between the children. This will maintain the flex spacing behavior and create a more straightforward implementation.

## Implementation

- Automated Spacing: The Gap widget should automatically insert appropriate spacing between the children of a flex container (Row or Column), adapting to the container's direction (horizontal or vertical).
- Directional Padding Application:
  - In horizontal layouts, add padding to the `right` of each child.
  - In vertical layouts, add padding to the `bottom` of each child.

Exclude padding for the last child to avoid unnecessary trailing space.

## Examples
CleanShot 2024-01-10 at 12 07 23

tilucasoli:
closes #150 

## Changes
* I changed the way the `Gap` works, now it's based on `Padding`
* Now we don't need any extra widget to do the gap feature

## Additional comments
* I followed the idea suggested by Leo on issue #150


lib/src/specs/flex/flex_widget.dart
```diff
@@ -1,11 +1,6 @@
 import 'package:flutter/widgets.dart';
 
-import '../../core/styled_widget.dart';
-import '../../factory/mix_provider.dart';
-import '../../factory/mix_provider_data.dart';
-import '../../widgets/gap_widget.dart';
-import '../container/box_widget.dart';
-import 'flex_spec.dart';
+import '../../../mix.dart';
```
leoafarias, line 3: This is normal to happen on `VsCode`, but it's important that we don't import from the mix and keep all the relative imports. I wonder if there is a configuration to make this default

leoafarias (changes requested):

lib/src/specs/flex/flex_widget.dart
tilucasoli: Done 👍

PR #159, opened 2024-01-18
Feature/styled icon supporting decorators

### LINKED ISSUE #156: StyledIcon doesn't support decorators attributes
### Package version

branch main and v0.7 too

### Flutter version

```bash
Flutter 3.16.5 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 78666c8dc5 (4 weeks ago) • 2023-12-19 16:14:14 -0800
Engine • revision 3f3e560236
Tools • Dart 3.2.3 • DevTools 2.28.4
```
### Steps to reproduce

1. Pass the `Style` as a parameter in `StyledIcon` constructor
2. In the `Style` put any attribute that is a decorator

### Expected results

The expected result is the `StyledIcon` receive all attributes and layout it.

### Actual results

The StyledIcon doesn't receive nothing that isn't related to Icon attributes

### Code sample

```dart
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
    return MaterialApp(
      home: StyledIcon(
        Icons.key,
        style: Style(
          icon.color.red(),
          icon.size(100),
          visibility(false),
        ),
      ),
    );
  }
}

```

tilucasoli:
#### Issue ticket number
closes #156 

## Describe your changes
* Create test cases for `Box` and `StyledBox`
* Add RenderWidgetDecorators in MixedIcon widget tree, so now you can pass decorators to StyledIcon
## Type of change
- **Bug fix** (non-breaking change which fixes an issue)
- **Test** (adding missing tests, refactoring tests; no production code change)

- [ ] I have made corresponding changes to the documentation

### DIFF
--- a/lib/src/specs/icon/icon_widget.dart
+++ b/lib/src/specs/icon/icon_widget.dart
@@ -1,9 +1,9 @@
 import 'package:flutter/material.dart';
 
 import '../../core/styled_widget.dart';
+import '../../decorators/widget_decorator_widget.dart';
 import '../../factory/mix_provider.dart';
 import '../../factory/mix_provider_data.dart';
-import '../container/box_widget.dart';
 import 'icon_attribute.dart';
 import 'icon_spec.dart';
 
@@ -24,14 +24,10 @@ class StyledIcon extends StyledWidget {
   @override
   Widget build(BuildContext context) {
     return withMix(context, (mix) {
-      return Box(
-        style: style,
-        inherit: inherit,
-        child: MixedIcon(
-          icon,
-          semanticLabel: semanticLabel,
-          textDirection: textDirection,
-        ),
+      return MixedIcon(
+        icon,
+        semanticLabel: semanticLabel,
+        textDirection: textDirection,
       );
     });
   }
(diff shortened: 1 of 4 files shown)


lib/src/decorators/widget_decorator_widget.dart
```diff
@@ -62,6 +62,9 @@ class RenderWidgetDecorators extends StatelessWidget {
       final decorator = decoratorMap.remove(decoratorType);
       if (decorator == null) continue;
       current = decorator.build(mix, current);
+
+      // Remove the decorator from the MixData, because it can be applied only once.
+      mix.removeType(decoratorType);
```
leoafarias, line 67: @tilucasoli I believe we might have to take another approach than mutating the attributes of the MixData; I think we might have to create an immutable class and pass it down to the child.

leoafarias (changes requested):

lib/src/decorators/widget_decorator_widget.dart
tilucasoli: Thanks, it's really important to remember! 
Thinking about possible solutions, I identified two. The first one is to propagate a new value in `RenderWidgetDecorators` using `MixProvider`. The second possibility is to create a new `MixData` without the decorator attributes when the existing `MixData` is inherited from `MixProvider`. My favorite solution is the second one, because adding a new widget to the widget tree unnecessarily doesn't seem like a good ideia.

lib/src/specs/icon/icon_widget.dart
```diff
@@ -1,11 +1,6 @@
 import 'package:flutter/material.dart';
 
-import '../../core/styled_widget.dart';
-import '../../decorators/widget_decorator_widget.dart';
-import '../../factory/mix_provider.dart';
-import '../../factory/mix_provider_data.dart';
-import 'icon_attribute.dart';
-import 'icon_spec.dart';
+import '../../../mix.dart';
```
leoafarias: Just be aware of this, vscode does this all the time to consolidate the exports. I wonder if there is a way to make sure it does not use mix.dart as an import

leoafarias (changes requested):

leoafarias (approved):
LGTM

Issue #175, opened 2024-01-27
Add ability to set PressableBox to disabled


JakesMD:
### Use case

Currently we're unable to set the `PressableBox` to disabled. Diving into the code for `PressableBox` we see `Pressable` has a `disabled` flag which isn't set (`false` by default)

### Proposal

Add an `isEnabled` flag to `PressableBox` which defaults to `true`:
``` dart
PressableBox(
  isEnabled: false,
  child: ...,
);
```

(I chose `isEnabled` here because I find it better not to use the negation. But could be `isDisabled` or `disabled` or whatever)

Alternatively the box could be set to disabled if `onPressed` is `null`.

Keep up the good work 🥳

tilucasoli:
Hello @JakesMD, thanks for your contribution! We appreciate it 😁.
We have developed a solution for your suggestion, please feel free to see it and make comments.

PR #183, opened 2024-01-29
Feat/issue 175

tilucasoli:
##### Issue ticket number
#175 

## Describe your changes
- Add a new attribute to PressableBox, `isDisabled`;
- Fix `isDisabled` feature in `Pressable` Widget
- Add more test cases

## Type of change
- **Bug fix** (non-breaking change which fixes an issue)


lib/src/widgets/pressable/pressable_widget.dart
```diff
@@ -72,7 +75,7 @@ class Pressable extends StatefulWidget {
   final VoidCallback? onLongPress;
   final FocusNode? focusNode;
   final bool autofocus;
-  final bool disabled;
+  final bool isDisabled;
```
leoafarias, line 78: I think we can keep `disabled` instead of `isDisabled`

lib/src/widgets/pressable/pressable_widget.dart
```diff
@@ -157,35 +160,48 @@ class PressableWidgetState extends State<Pressable> {
         button: true,
         focusable: onEnabled && _node.canRequestFocus,
         focused: _node.hasFocus,
-        child: GestureDetector(
```
leoafarias, line 160: I think the disabled should behave only for the onTap behavior, we also should not modify the element in the tree through the function returning a widget.

leoafarias (changes requested):

lib/src/widgets/pressable/pressable_widget.dart
tilucasoli: Done 👍

lib/src/widgets/pressable/pressable_widget.dart
tilucasoli: I'll change the function. However, about the disabled state only affects `onTap`. Are you sure? In my opinion when disabled is `true`, all actions involving the `Pressable` should be deactivated.

leoafarias: @tilucasoli makes sense!

leoafarias (approved):
LGTM

PR #185, opened 2024-01-30
Create StyledImage

### LINKED ISSUE #166: The mix doesn't have a StyledImage with its own attributes
### Use case

We have a bunch of `Widgets` from Mix, but there isn't a specific one to Image. 

### Proposal

The proposal is to create a new `Widget` called `StyledImage`. Its constructors will be as follows:
```dart
const StyledImage({
  Key? key,
  required ImageProvider<Object> image,
  Style style,
  ImageFrameBuilder? frameBuilder,
  ImageLoadingBuilder? loadingBuilder,
  ImageErrorWidgetBuilder? errorBuilder,
  String? semanticLabel,
  bool excludeFromSemantics = false,
})
```
The `StyledImage` will be designed to accept the following attributes via the `Style`:

- `double` width (Done)
- `double` height (Done)
- `color` (Done)
- `double` opacity (Done)
- `BoxFit` fit (Done)
- `AlignmentGeometry` alignment (Done)
- `ImageRepeat` repeat
- `Rect` centerSlice
- `FilterQuality` filterQuality
-  `BlendMode` colorBlendMode

tilucasoli:
#### Issue ticket number
resolve #166

## Describe your changes
- Create StyledImage and a doc for it;
- Implement new utilities, `RectUtility` and `FilterQualityUtility`;
- Create new test cases;
- Add `test_resources` to test Image widget.

## Type of change
- **Test** (adding missing tests, refactoring tests; no production code change)


lib/src/specs/image/image_spec.dart
```diff
@@ -43,6 +54,10 @@ class ImageSpec extends Spec<ImageSpec> {
       color: Color.lerp(color, other?.color, t),
       repeat: t < 0.5 ? repeat : other?.repeat,
       fit: t < 0.5 ? fit : other?.fit,
+      centerSlice: Rect.lerp(centerSlice, other?.centerSlice, t),
+      filterQuality: t < 0.5 ? filterQuality : other?.filterQuality,
```
leoafarias, line 58: We have a utility called `lerpSnap` to handle this type of lerp transition

lib/src/specs/image/image_util.dart
```diff
@@ -45,4 +53,21 @@ class ImageUtility extends SpecUtility<ImageSpecAttribute> {
...
+  }
+
+  FilterQualityUtility<ImageSpecAttribute> get filterQuality {
+    return FilterQualityUtility(
+        (filterQuality) => _only(filterQuality: filterQuality));
+  }
+
+  BlendModeUtility<ImageSpecAttribute> get blendMode {
+    return BlendModeUtility((blendMode) => _only(blendMode: blendMode));
+  }
```
leoafarias, line 72: Let's also add a `call()` method. One to keep in mind, though, is the call() method takes the actual Flutter core classes, not Dtos, if they exist.

test/src/specs/image/image_attribute_test.dart
```diff
@@ -52,18 +52,26 @@ void main() {
...
         fit: BoxFit.cover,
+        alignment: Alignment.bottomCenter,
+        centerSlice: Rect.zero,
+        filterQuality: FilterQuality.low,
+        colorBlendMode: BlendMode.srcOver,
       );
       final props = attribute.props;
 
-      expect(props.length, 5);
+      expect(props.length, 9);
```
leoafarias, line 65: I think we don;t need to do these props check, as they are hard to maintain, and I think some `dcm` fixes might reorder them. The most important thing is to check the values, and also check `equality` on the tests.

test/src/specs/image/image_widget_test.dart
```diff
@@ -0,0 +1,88 @@
...
+    testWidgets('receive all the attributes', (WidgetTester tester) async {
+      await tester.pumpWidget(
+        MaterialApp(
+          home: Scaffold(
+            body: StyledImage(
+              style: Style(
+                image.width(152),
+                image.height(152),
+                image.color.black(),
+                image.repeat.repeat(),
```
leoafarias, line 18: These cases are always a bit tough, but they seem redundant, can we have the call method on `image.repeat()` be ImageRepeat.repeat?

leoafarias (changes requested):
Looking really good!

lib/src/specs/image/image_util.dart
```diff
@@ -16,21 +16,29 @@ class ImageUtility extends SpecUtility<ImageSpecAttribute> {
...
+      filterQuality: filterQuality,
     );
   }
 
   ColorUtility<ImageSpecAttribute> get color {
     return ColorUtility((color) => _only(color: color));
   }
 
-  ImageRepeatUtility<ImageSpecAttribute> get repeat {
+  ImageRepeatUtility<ImageSpecAttribute> get imageRepeat {
```
leoafarias, line 41: This would still be `image.imageRepeat()`, what I suggest is what you had before, and try to remove redundant words. so we would keep `repeat,` but add a callable method to return `ImageRepeat.repeat`

leoafarias (changes requested):

lib/src/specs/image/image_util.dart
tilucasoli: Sorry, my bad 😅

leoafarias (changes requested):

leoafarias (approved):
LGTM

PR #205, opened 2024-02-09
Implementing Style Builder

leoafarias:
## Describe your changes

Create style builder for dynamic creation of styling attributes

## Type of change

- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works


lib/src/specs/container/box_util.dart
```diff
@@ -286,6 +286,8 @@ final clipBehavior = box.clipBehavior;
 /// - [BoxShadow]
 final elevation = box.elevation;
 
+final radialGradient = box.decoration.gradient.radial;
```
tilucasoli, line 289: In the most part of the best practices and Remix, we are using box.{some attribute} so I think we could follow this pattern.

PR #214, opened 2024-02-23
Feat/add deprecated apis

tilucasoli:
## Describe your changes
- I added more deprecated cases to facilitate smoother package update

## Type of change


tilucasoli:
We have two edge cases. The first one is the `animated` attribute, for which we don't have a direct comparison. Perhaps the `PressableBox` component can help, especially in XP usage. The other case is the `align` attribute, which used to refer to `Alignment`, but is now a decorator, resulting in an API change. In any case, I'll create two new sections in the migration guide to address them.

PR #227, opened 2024-03-15
Create documentation for Theming

tilucasoli:
## Describe your changes
- Create a new Doc

## Type of change

- [ ] I have performed a self-review of my code
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works


leoafarias:
@tilucasoli I think we should also describe the implementation of the  Color tokens on the ColorUtil as extension, so something like `backgroundColor.primary()` for example can be used.

PR #228, opened 2024-03-15
Clean up design tokens pre implemented

tilucasoli:
## Describe your changes
- Remove pre-implemented Space and Radius Tokens
- Remove tokens utilities for Space and Radius
- Rename space to spaces
## Type of change

- **New feature**
- **Test**

### DIFF
--- a/lib/src/theme/mix/mix_tokens.dart
+++ b/lib/src/theme/mix/mix_tokens.dart
@@ -1,67 +1,14 @@
 import 'package:meta/meta.dart';
 
-import '../tokens/color_token.dart';
 import '../tokens/radius_token.dart';
 import '../tokens/space_token.dart';
-import '../tokens/text_style_token.dart';
 
 @immutable
 class MixTokens {
-  final color = const _MixColorTokens();
-  final textStyle = const _MixTextStyles();
   final space = const _MixSpaces();
   final radius = const _MixRadii();
 
   const MixTokens();
-}
-
-@immutable
-class _MixColorTokens {
-  final primary = const ColorToken('mix.color.primary');
-  final secondary = const ColorToken('mix.color.secondary');
-  final tertiary = const ColorToken('mix.color.tertiary');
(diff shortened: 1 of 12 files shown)


lib/src/theme/mix/mix_tokens.dart
```diff
@@ -0,0 +1,83 @@
...
+  final textStyle = const _MixTextStyles();
+  final space = const _MixSpaces();
+  final radius = const _MixRadii();
+
+  const MixTokens();
+}
+
+@immutable
+class _MixColorTokens {
+  final primary = const ColorToken('mix.color.primary');
```
leoafarias, line 20: These would not be Mix Tokens but Material ones. I don't think we should use them as MixTokens by default but maintain the ability to import material tokens into it.

leoafarias (changes requested):

lib/src/theme/tokens/radius_token.dart
```diff
@@ -58,27 +50,3 @@ class RadiusRef extends Radius with TokenRef<RadiusToken, Radius> {
   @override
   int get hashCode => token.hashCode;
 }
-
-// // Helper class to wrap functions that can return
-// a Radius token that resmebles the WithSpaceToken
-@immutable
-class UtilityWithRadiusTokens<T> {
```
leoafarias, line 65: Are the other utilities still within Mix? I would like to maintain this here, to help compose utilities.

leoafarias (changes requested):
LGTM

lib/src/theme/tokens/radius_token.dart
tilucasoli: Leo, are you sure about this? This class wasn't instantiated anywhere in the code.

lib/src/theme/mix/mix_tokens.dart
tilucasoli: done

PR #250, opened 2024-04-25
Update Variant documentation

tilucasoli:
## Describe your changes
- Update variant documentation
## Type of change

### DIFF
--- a/website/pages/docs/guides/variants.mdx
+++ b/website/pages/docs/guides/variants.mdx
@@ -130,6 +130,16 @@ final style = Style(
 );
 ```
 
+## Prioritizing Variants
+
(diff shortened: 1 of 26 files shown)


website/pages/docs/guides/variants.mdx
```diff
@@ -130,6 +130,16 @@ final style = Style(
 );
 ```
 
+## Prioritizing Variants
```
leoafarias, line 133: Prioritization should be only on `ContextVariants` as the use of `Variants` are explicit in its priority.

leoafarias (changes requested):

website/pages/docs/guides/variants.mdx
tilucasoli: So should I remove the priority parameter in the Variants constructor?

tilucasoli: I don't know if I'm sure about this. Imagine that we have a variant, such as disabled in a button. It's a normal variant that needs priority if compared with other ones.

PR #332, opened 2024-06-26
fix: callable specs and modifiers

tilucasoli:
### Description
Some fixes in call methods and modifiers structure

### Changes

- Make `WithModifierUtility` a generic class and receive a builder in its constructor;
- Make callable specs have the same parameters as the spec widget constructor;
- Fix `PaddingModifierAttribute` merge.

**Review Checklist**

### Impacted Packages

Please indicate which packages in our project are affected by these changes:

- [x] mix
- [ ] mix_lint
- [ ] mix_generator
- [ ] documentation

Looking at the `WithModifierUtility` changes, maybe it can be a breaking change in specific scenarios. What do you think @leoafarias?

### DIFF
--- a/packages/mix/lib/src/specs/text/text_spec.dart
+++ b/packages/mix/lib/src/specs/text/text_spec.dart
@@ -52,14 +52,21 @@ final class TextSpec extends Spec<TextSpec> with _$TextSpec {
     super.animated,
   });
 
-  Widget call(String text) {
+  Widget call(String text, {String semanticLabel = '', Locale? locale}) {
     return isAnimated
         ? AnimatedTextSpecWidget(
             text,
             spec: this,
+            semanticsLabel: semanticLabel,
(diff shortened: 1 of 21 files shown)


packages/mix/lib/src/specs/text/text_spec.dart
```diff
@@ -52,14 +52,21 @@ final class TextSpec extends Spec<TextSpec> with _$TextSpec {
     super.animated,
   });
 
-  Widget call(String text) {
+  Widget call(String text, {String semanticLabel = '', Locale? locale}) {
```
leoafarias, line 55: We should not pass defaults

leoafarias (changes requested):

leoafarias (approved):

leoafarias (approved):

Issue #313, opened 2024-06-20
Ability to add WidgetModifiers to other Specs using specUtility.mod


leoafarias:
Since `WidgetModifierSpec` is a spec with an attribute also, it would be good for the `mod` utility to accept any spec that has a build method that has a `SingleChildWidgetSpec`, as it would be the same signature as the `WidgetModifiers`.

Also, this would allow us to add the specs to the code generation, as it would bring regular specs and widget modifiers closer together without causing breaking changes.

In that case, the `WidgetModifierUtility` would have a call method that would accept a `SingleChildWidgetSpecAttribute`, which all modifiers accept, but also items like the `BoxSpecAttribute` would also be used, or any custom specs generated that accept a child as an argument.

Here are some examples of the signature.

```dart
abstract class WidgetSpec<T extends Spec<T>> extends Spec<T> {
  const WidgetSpec({super.animated});
  Widget build();
}

abstract class SingleChildWidgetSpec<T extends Spec<T>> extends WidgetSpec<T> {
  @override
  Widget build({Widget? child});
}

abstract class MultiChildWidgetSpec<T extends Spec<T>> extends WidgetSpec<T> {
  @override
  Widget build({List<Widget>? children});
}
```

PR #333, opened 2024-06-27
feat: modifiers in spec

tilucasoli:
#### Related issue
#313 

### Description
Add the ability to set modifiers in **specs**.

### Changes

- Add modifiers as an attribute of `Spec`;
- Add the ability to set modifiers in **specs**;
- Change the default way to reorder modifiers, now it should be a `Spec` instead of `Attribute`;
- Deprecate some attributes of `RenderModifiers`;
- Create a directory, which will group the old tests to ensure doesn't exist breaking changes.

**Review Checklist**

I tried not to create breaking changes, let me know if you have any doubts about some modifications.


packages/mix/lib/src/specs/icon/icon_widget.dart
```diff
@@ -50,28 +51,32 @@ class IconSpecWidget extends StatelessWidget {
     this.semanticLabel,
     super.key,
     this.textDirection,
-    this.modifierOrder = const [],
+    this.orderOfModifiers = const [],
   });
 
   final IconData? icon;
   final IconSpec? spec;
```
tilucasoli, line 58: For the Widget spec, all of them follow the same basic structure, a spec attribute with some `Spec` and an attribute called orderOfModifiers. What if we create an `abstract class SpecWidget`?

packages/mix/lib/src/specs/icon/icon_widget.dart
```diff
@@ -117,7 +122,7 @@ class AnimatedIconSpecWidget extends ImplicitlyAnimatedWidget {
     super.key,
     this.semanticLabel,
     this.textDirection,
-    this.modifierOrder = const [],
+    this.orderOfModifiers = const [],
```
tilucasoli, line 125: it will cause break changes

packages/mix/lib/src/specs/box/box_spec.dart
```diff
@@ -96,17 +96,22 @@ final class BoxSpec extends Spec<BoxSpec> with _$BoxSpec {
     this.clipBehavior,
     this.width,
     this.height,
+    super.modifiers,
     super.animated,
   });
 
-  Widget call({Widget? child}) {
+  Widget call({Widget? child, List<Type> orderOfModifiers = const []}) {
```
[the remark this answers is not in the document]
tilucasoli: @leoafarias if it doesn't have a default value, it will cause a breaking change.

tilucasoli:
@leoafarias do you think we should create a document for this migration now?

PR #345, opened 2024-07-02
fix: Exception when there is no children on flex

leoafarias:
Exception on flex widget when there are no children and it tries to add a gap.

**Review Checklist**

- [ ] **Testing**: Have you tested your changes, including unit tests and integration tests for affected code?

### Impacted Packages

Please indicate which packages in our project are affected by these changes:

- [x] mix
- [ ] mix_lint
- [ ] mix_generator
- [ ] documentation

### DIFF
--- a/packages/mix/lib/src/specs/flex/flex_widget.dart
+++ b/packages/mix/lib/src/specs/flex/flex_widget.dart
@@ -2,7 +2,6 @@
 
 import 'package:flutter/widgets.dart';
 
-import '../../core/factory/style_mix.dart';
 import '../../core/styled_widget.dart';
 import '../box/box_spec.dart';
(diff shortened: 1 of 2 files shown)


packages/mix/lib/src/specs/flex/flex_widget.dart
```diff
@@ -211,11 +212,11 @@ class FlexBox extends StyledWidget {
 /// ```
 class HBox extends FlexBox {
```
tilucasoli, line 213: What do you think about making a test for this specific use case?

tilucasoli (changes requested):

tilucasoli (approved):

PR #374, opened 2024-07-11
refactor: use timer instead of future<void>

tilucasoli:
### Description

Use timer instead of future<void>. Because of this, we can dispose the timer.

### Changes

- Add a timer in `_MixStateWidgetBuilderState`

**Review Checklist**

### DIFF
--- a/packages/mix/lib/src/widgets/widget_state/pressable_widget.dart
+++ b/packages/mix/lib/src/widgets/widget_state/pressable_widget.dart
@@ -294,6 +294,9 @@ abstract class _MixStateWidgetBuilderState<T extends _MixStateWidgetBuilder>
       }
     }
 
+    _timer?.cancel();
+    _timer = null;
+
     final delay = widget.unpressDelay;
 
     if (delay != Duration.zero) {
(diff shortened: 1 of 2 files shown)


packages/mix/lib/src/widgets/widget_state/pressable_widget.dart
```diff
@@ -277,15 +279,27 @@ abstract class _MixStateWidgetBuilderState<T extends _MixStateWidgetBuilder>
...
+        updateState(() => _isPressed = false);
+      }
     }
 
-    if (_isPressed && _pressCount == initialPressCount) {
-      updateState(() => _isPressed = false);
+    final delay = widget.unpressDelay;
+
+    if (delay != Duration.zero) {
+      _timer = Timer(delay, unpressCallback);
```
leoafarias, line 300: Look into canceling the timer before setting a new one. Not sure if it applies to this.

tilucasoli: Done 👍🏻

PR #379, opened 2024-07-15
fix: add flexible as the first in the order of modifiers

### LINKED ISSUE #373: Flexible and Align modifiers doesn't work together
### Package version

1.2.0

### Flutter version

3.22.2

### Steps to reproduce

Run the code sample

### Expected results

Box in Align in Flexible. Flexible(child: Align(child: Box()))

### Actual results

The `Incorrect use of ParentDataWidget` error because the Align modifier goes before the Flexible even though it's placed first. That's because of the modifiers default order: Align modifier is listed there but Flexible is not so Flexible is sorted to the end of the modifiers list. 

By the way, I found it very misleading that there is a default modifiers order. It's not mentioned anywhere in docs. And I think the default order it's not always good. For example, what if I want to set the box size and then align it's content? To do so I can't use only `$with` or only `$box`, I must use the `$with.sizedBox` and then `$box.alignment`. `$box.height` and `$with.align` won't work!

In my humble opinion, mix shouldn't have the default ordering. The list of modifiers provided by the developer should be the source of truth. And I don't want to override or disable default ordering all the time! 🙂

Full error message:
```
The ParentDataWidget Flexible(flex: 1) wants to apply ParentData of type FlexParentData to a RenderObject, which has been set up to accept ParentData of incompatible type BoxParentData.
Usually, this means that the Flexible widget has the wrong ancestor RenderObjectWidget. Typically, Flexible widgets are placed directly inside Flex widgets.
The offending Flexible is currently placed inside a Align widget.
The ownership chain for the RenderObject that received the incompatible parent data was:
  SizedBox ← RenderModifiers ← Mix ← Box ← Container ← BoxSpecWidget ← Builder ← Flexible ← Align ← RenderModifiers ← ⋯
```

### Code sample

<details open><summary>Code sample</summary>

```dart
VBox(
  children: [
    Box(
      style: Style(
        $with.expanded(),
        $with.align(alignment: Alignment.center),
      ),
      child: Box(
        style: Style(
          $with.sizedBox.square(50),
          $box.color.red(),
        ),
      ),
    ),
  ],
)
```

</details>

tilucasoli:
#### Related issue
closes #373 

### Description

Review the order of modifiers

### Changes

Add FlexibleModifier, PaddingModifier, and RotatedModifier.

**Review Checklist**

- [ ] **Testing**: Have you tested your changes, including unit tests and integration tests for the affected code?

### DIFF
--- a/packages/mix/lib/src/modifiers/render_widget_modifier.dart
+++ b/packages/mix/lib/src/modifiers/render_widget_modifier.dart
@@ -5,46 +5,43 @@ import 'package:flutter/material.dart';
 import '../core/factory/mix_data.dart';
 import '../core/modifier.dart';
 import '../core/spec.dart';
-import 'align_widget_modifier.dart';
-import 'aspect_ratio_widget_modifier.dart';
-import 'clip_widget_modifier.dart';
-import 'fractionally_sized_box_widget_modifier.dart';
-import 'intrinsic_widget_modifier.dart';
-import 'opacity_widget_modifier.dart';
-import 'sized_box_widget_modifier.dart';
-import 'transform_widget_modifier.dart';
-import 'visibility_widget_modifier.dart';
+import 'modifiers.dart';
 
 const _defaultOrder = [
-  // 1. VisibilityModifier: Controls overall visibility. If the widget is set to be invisible,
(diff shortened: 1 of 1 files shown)


leoafarias (approved):
LGTM

PR #414, opened 2024-07-30
fix: FlexSpecWidget prioritizes the direction in spec

tilucasoli:
### Description

This PR makes `FlexSpecWidget` prioritize the direction specified in the `FlexSpec`.

### Changes

- Prioritize `direction` in `FlexSpec` instead of the parameter.

- This PR changes this behavior in all `Flex` alternatives.


leoafarias (approved):
LGTM

Issue #349, opened 2024-07-04
`SingleChildScrollView` modifier


Maksimka101:
### Use case

V and H boxes are pretty powerful, but sometimes I need to make them scrollable 

### Proposal

Add something like `$with.scrollable` modifier on base of the `SingleChildScrollView` to make anything scrollable. Probably it can even inherit axis of the FlexBox

Maksimka101:
Btw, I'd love to implement it

tilucasoli:
Thank you for your contribution! 😁

It's an interesting use case of `modifiers` and mix. Your suggestion is between a behavior and a layout modification, so we should be careful if it would be part of Mix or an internal `modifier` in your project. We'll work on some proof of concepts related to it and we'll keep you up to date.

dickermoshe:
Close this

PR #427, opened 2024-08-03
feat: `SingleChildScrollView` widget modifier

Maksimka101:
#### Related issue

#349 

### Description

Added the `SingleChildScrollView` widget modifier

### Changes

List specific changes made in this PR.

**Review Checklist**

Is there any additional context or documentation that might be helpful for reviewers?


Maksimka101:
Btw, why are all `*SpecUtility` classes are final?

packages/mix/lib/src/modifiers/scroll_view_widget_modifier.dart
```diff
@@ -0,0 +1,145 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/widgets.dart';
+import 'package:mix/mix.dart';
+import 'package:mix_annotations/mix_annotations.dart';
+
+part 'scroll_view_widget_modifier.g.dart';
+
+@MixableSpec(skipUtility: true, withLerp: false)
+final class ScrollViewModifierSpec
```
tilucasoli, line 10: We acknowledge that there are numerous customizations possible with the `SingleChildScrollView`. However, supporting all of them would compromise the architecture. The `Attributes` were intended for visual styles, so if we incorporate properties like controllers, it would begin to manage behavior that is not its intended function.

.vscode/settings.json
```diff
@@ -5,6 +5,7 @@
     "source.dcm.fixAll": "explicit"
   },
   "dart.flutterSdkPath": ".fvm/versions/3.19.0",
+  "dart.lineLength": 80,
```
tilucasoli, line 8: We understand it's a personal option. However, we are inclined to go with the previous option.

tilucasoli (approved):

tilucasoli (changes requested):
Thank you for your contribution ❤️.

To fine-tune this feature we still need to brainstorm which will be handled by this modifier and what kind of properties it will support

packages/mix/lib/src/modifiers/scroll_view_widget_modifier.dart
tilucasoli: based on this comment above, this modifier shouldn't be a complete API to become widgets scrollable. It's more like a simple way to handle simple cases.

tilucasoli (changes requested):

.vscode/settings.json
Maksimka101: pub.dev requires such line length. Why not add this rule to have correct formatting without a separate command?

Maksimka101:
I've removed all non visual properties except `primary` as it feels useful one

Should I change anything else?

packages/mix/lib/src/modifiers/scroll_view_widget_modifier.dart
```diff
@@ -0,0 +1,104 @@
...
+  T direction(Axis axis) => call(scrollDirection: axis);
+
+  /// Set the scroll direction of the scroll view to horizontal.
+  T horizontal() => call(scrollDirection: Axis.horizontal);
+
+  /// Set the scroll direction of the scroll view to vertical.
+  T vertical() => call(scrollDirection: Axis.vertical);
+
+  /// Make the scroll view reverse or not.
+  T reverse([bool reverse = true]) => call(reverse: reverse);
```
tilucasoli, line 62: What do you think about use `BoolUtility`
```suggestion
 BoolUtility<T> get reverse => BoolUtility((v) => call(reverse: v));
```

tilucasoli:
LGTM!

tilucasoli (approved):

tilucasoli (approved):

PR #466, opened 2024-08-23
feat: Fold MixThemeData into single root

Solido:
There are two ways to do that.

``` dart
instance.fold([themes...]) // require an existing instance.
MixThemeData.fold([themes...]) // auto instanciation
```

I found the second less intrusive but we can have both 
we just need to decide on good naming.

### DIFF
--- a/packages/mix/lib/src/theme/mix/mix_theme.dart
+++ b/packages/mix/lib/src/theme/mix/mix_theme.dart
@@ -132,6 +132,13 @@ class MixThemeData {
     );
   }
 
+  /// Merge all [themes] into a single [MixThemeData] root.
+  static MixThemeData fold(Iterable<MixThemeData> themes) {
+    if (themes.isEmpty) return const MixThemeData.empty();
+    return themes.fold(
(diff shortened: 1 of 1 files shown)


packages/mix/lib/src/theme/mix/mix_theme.dart
```diff
@@ -132,6 +132,13 @@ class MixThemeData {
     );
   }
 
+  /// Merge all [themes] into a single [MixThemeData] root.
+  static MixThemeData fold(Iterable<MixThemeData> themes) {
```
tilucasoli, line 136: The implementation is perfect. The only thing I'm considering is using the name 'merge' instead of 'fold' for this method for better readability.

tilucasoli: @leoafarias

leoafarias: I agree with the initial comment. I think we have similar behavior for styles, and we use the static `combine`. I believe that is consistent and explains it well.

PR #471, opened 2024-08-27
fix: Update OnBrightnessVariant to use `MediaQuery` instead of `Theme`

doug-orchard:
Replaced Material specific detection of brightness with a generic brightness detector.

### Description
Theme.of(context) is Material dependant. When using Cupertino, $on.dark breaks.
Replaced the brightness detection from Material to MediaQuery.

### Changes

• Updated brightness detection to MediaQuery.of() from Theme.of()

**Review Checklist**

- [ ] **Testing**: Have you tested your changes, including unit tests and integration tests for affected code?

Is there any additional context or documentation that might be helpful for reviewers?


packages/mix/lib/src/variants/context_variant_util/on_brightness_util.dart
```diff
@@ -21,7 +21,7 @@ class OnBrightnessVariant extends MediaQueryContextVariant {
   /// [brightness] associated with this variant, and `false` otherwise.
   @override
   bool when(BuildContext context) {
-    return Theme.of(context).brightness == brightness;
+    return MediaQuery.of(context).platformBrightness == brightness;
```
leoafarias, line 24: Can we use the InheritedModel method of MediaQuery?
`MediaQuery.platformBrightnessOf(context) == brightness;`

leoafarias (changes requested):
One small change: thanks for the fix. It seems the tests are also failing, and we will have to update them. Let me know if you can take a look at that; if not, I can.

packages/mix/lib/src/variants/context_variant_util/on_brightness_util.dart
doug-orchard: The main issue is the tests are focused on the brightness that attached to the apps theme. MediaQuery pulls from the device. Hence the tests failing. 

The goal is for $on.dark to work for Cupertino apps. Not 100% sure this ( MediaQuery ) is the right path.

leoafarias: @doug-orchard The `platformBrightnessOf` will not change the behavior of your PR; it will only watch the `platformBrightness` value instead of any `MediaQuery` change.

doug-orchard: Okay! That sounds great.

Time and knowledge of Mix is holding me back to delve deeper that this change required. So, @leoafarias, feel free to take over.

PR #577, opened 2025-04-01
docs: Create two examples of usage for mix

tilucasoli:
### Description

Create two examples of usage for mix:
- One animated component
- One Design System Example


tilucasoli:
We will wait for the mix 2.0 to release more complete examples

PR #631, opened 2025-06-09
refactor: Replace gap with spacing

tilucasoli:
### Description

This PR refactors the codebase by replacing the `gap` attribute with the new `spacing` property across theme tokens, Flex widgets, specs, and utilities.
### Changes

Key changes include updating extension types to use `FlexSpacingUtility`, modifying tests to assert spacing on `Flex` widgets, and deprecating legacy `gap` fields and helper methods.

**Review Checklist**

- [x] **Breaking Changes**: Does this change introduce breaking changes affecting existing code or users?

### DIFF
--- a/packages/mix/lib/src/specs/flex/flex_spec.dart
+++ b/packages/mix/lib/src/specs/flex/flex_spec.dart
@@ -41,11 +41,20 @@ final class FlexSpec extends Spec<FlexSpec> with _$FlexSpec, Diagnosticable {
   final TextBaseline? textBaseline;
   final Clip? clipBehavior;
 
+  @Deprecated(
+    'Use spacing instead. '
+    'The gap property has been replaced with spacing for better naming consistency. '
+    'This property will be removed in v2.0.0.\n',
+  )
   @MixableField(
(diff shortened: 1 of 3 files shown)


packages/mix/lib/src/specs/flex/flex_spec.dart
leoafarias, line 1: @tilucasoli I would prefer not to break the code generation on this since it's an important piece. Maybe we need to define a bit more the paint point and issues as well, and keep in mind that the Flex widget now does have a gap property for the new version of Flutter.

tilucasoli: In this pull request, I am replacing the `gap` property with `spacing` to align it with Flutter's implementation. An alternative approach would be to remove our current implementation and have `gap` use the spacing implementation without making any changes to the Utility, Attribute, or Specification.

tilucasoli: @leoafarias 

https://github.com/btwld/mix/pull/632

PR #648, opened 2025-06-13
fix: Change Mixable to extend EqualityMixin instead of StyleElement

tilucasoli:
## Summary
• Remove StyleElement inheritance from Mixable class
• Add EqualityMixin with mergeKey implementation  
• Maintain existing resolve and merge method signatures

## Test plan
- [ ] Verify existing tests pass
- [ ] Check that Mixable classes still function correctly
- [ ] Ensure merge behavior is preserved

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/mix/lib/src/core/element.dart
+++ b/packages/mix/lib/src/core/element.dart
@@ -28,8 +28,6 @@ typedef Dto<Value> = Mixable<Value>;
 abstract class Mixable<Value> with EqualityMixin {
   const Mixable();
 
-  Object get mergeKey => runtimeType;
-
   Value resolve(MixData mix);
   Mixable<Value> merge(covariant Mixable<Value>? other);
 }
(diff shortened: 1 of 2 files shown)


packages/mix/lib/src/core/element.dart
```diff
@@ -25,12 +25,12 @@ typedef StyledAttribute = SpecAttribute;
 @Deprecated('Use Mixable instead')
 typedef Dto<Value> = Mixable<Value>;
 
-abstract class Mixable<Value> extends StyleElement {
+abstract class Mixable<Value> with EqualityMixin {
   const Mixable();
 
-  Value resolve(MixData mix);
+  Object get mergeKey => runtimeType;
```
leoafarias: I think we can remove the merge key from mixable since the style element already has it.

leoafarias (changes requested):

packages/mix/lib/src/core/element.dart
tilucasoli: Done

leoafarias (approved):
LGTM

PR #708, opened 2025-07-06
Feature/type registry autodiscovery

leoafarias:
Improved self-discovery system with the use of the `gen_helpers` package in preparation for 2.0 changes.


leoafarias:
Closing — the core type registry autodiscovery infrastructure has already landed on main through subsequent work. This 8-month-old PR (151 files) has heavy merge conflicts, and the remaining value (validation tooling, docs, DTO consolidation) would need a fresh approach against the current generator architecture.

PR #853, opened 2026-02-24
feat(mix): add MixScope.inherit to combine parent and child tokens

tilucasoli:
### Description

Adds **MixScope.inherit**, a way to combine tokens from the nearest parent MixScope with additional tokens in a single merged scope. A nested `MixScope` normally replaces the parent for its subtree; `MixScope.inherit` instead merges parent + child token maps so both upstream and local tokens resolve. Inner tokens override parent when the same token key is defined in both. Implemented via a Builder that reads the parent scope and builds one MixScope with the merged map (no change to `getToken`/`tokenOf` resolution).

### Changes

- **packages/mix/lib/src/theme/mix_theme.dart**
  - Add static `MixScope.inherit()` that takes the same token maps as the main factory, uses a Builder to get `MixScope.maybeOf(context, 'tokens')`, merges `{...?parent?.tokens, ...childTokens}`, and returns a single `MixScope` with merged tokens and optional `orderOfModifiers` (child then parent).
- **packages/mix/test/src/theme/mix_scope_inherit_test.dart**
  - New tests: outer + inner tokens both resolve; Material (Fortal-like) + inherit; resolution for `.light` and `.dark`; inner overrides parent when same key; inherit with no parent.
- **packages/mix/doc/mix-scope-and-theming.md**
  - Note that nested MixScope replaces parent; document `MixScope.inherit` as the way to combine tokens (parent + yours, yours override). Example with plain outer `MixScope` and `MixScope.inherit`.

**Review Checklist**

- [x] **Testing**: Unit/integration tests added for inherit (outer+inner, Material+inherit, light/dark, override, no parent).
- [ ] **Breaking Changes**: No. New API only.
- [x] **Documentation Updates**: mix-scope-and-theming.md updated.
- [ ] **Website Updates**: N/A unless docs are mirrored on the website.

Validation (from original what-if): one “Fortal” token and one custom Mix token both resolve in the same subtree; resolution validated for `.light` and `.dark` within the outer scope.


packages/mix/lib/src/theme/mix_theme.dart
```diff
@@ -66,6 +66,65 @@ class MixScope extends InheritedModel<String> {
...
+      tokens: childTokens,
+      orderOfModifiers: orderOfModifiers,
+      child: child,
+    );
+
+    return Builder(
+      builder: (context) {
+        final parent = MixScope.maybeOf(context);
+
+        return MixScope.combine(scopes: [?parent, nestedScope], child: child);
```
leoafarias: **Bug: `key` parameter is silently ignored**

The `key` passed to `inherit()` goes into `nestedScope`, but `nestedScope` is never inserted into the widget tree — it's only used as a data carrier for `combine()`. The `Builder` returned here doesn't receive the key, and `combine` is called without forwarding it (defaults to `null`).

A caller passing `key: ValueKey('foo')` would expect it to affect widget identity, but it has zero effect.

**Suggested fix** — forward `key` to `combine`:
```dart
return MixScope.combine(key: key, scopes: [?parent, nestedScope], child: child);
```

packages/mix/doc/mix-scope-and-theming.md
```diff
@@ -182,6 +182,20 @@ MixScope(
 )
 ```
 
+A nested `MixScope` is the *nearest* scope for its subtree, so it *replaces* the parent for token resolution. To **combine** tokens—keep the parent’s tokens and add (or override) your own—use **MixScope.inherit**. It builds a single scope whose token map is the parent’s map merged with yours: parent tokens plus your tokens, with your entries winning when the same token is defined in both. That way both upstream tokens and your local tokens resolve in the same subtree. Valid for both light and dark themes.
+
+```dart
+// Outer scope provides base tokens; inherit combines them with yours
+MixScope(
+  colors: { ColorToken('brand.primary'): Colors.blue },
```
leoafarias: **Doc: Invalid Dart — duplicate named parameter `colors:`**

This constructor call has `colors:` twice, which is a compile error in Dart. The second entry should be merged into a single map:

```dart
MixScope(
  colors: {
    ColorToken('brand.primary'): Colors.blue,
    ColorToken('custom.accent'): Colors.orange,
  },
  child: MixScope.inherit(
    spaces: { SpaceToken('custom.gap'): 12.0 },
    child: MySubtree(),
  ),
)
```

leoafarias (commented):
Code review — 2 issues found (details in inline comments below).

leoafarias (changes requested):

PR #921, opened 2026-05-28
chore(mix): default StyleWidget to IdentityStyle<Spec>

leoafarias:
## Summary
- Default `style` on every `StyleWidget<S>` (`Box`, `StyledIcon`, `FlexBox`, `StyledImage`, `Pressable`, `StackBox`, `StyledText`) is now `const IdentityStyle<S>(const S())` instead of the concrete `*Styler.create()`.
- Widget files no longer import their companion `*_style.dart`; they depend only on the `Spec` and the generic plumbing in `core/style.dart`.
- Supporting tweaks in `core/style.dart` and `core/style_builder.dart`; new tests in `test/src/core/identity_style_test.dart` and `style_widget_default_style_test.dart`.

## Motivation
Removing the widget → custom-Styler dependency is a prerequisite for code-generating widget files from the `Spec` alone. With this in place, the generator does not need to know which `*Styler` subclass exists for each spec.

## Test plan
- [x] `melos run gen:build`
- [x] `melos run ci` (test:flutter + test:dart — all green)
- [ ] `melos run analyze` — `mix` package is clean (`No issues found!`). `mix_tailwinds` reports pre-existing issues identical to `origin/main` (3777, unrelated to this branch).
- [x] Confirmed no `Styler.create()` remains as a widget-level default under `packages/mix/lib/src/specs/**/*_widget.dart`.

### DIFF
--- a/packages/mix/lib/src/core/style.dart
+++ b/packages/mix/lib/src/core/style.dart
@@ -64,26 +64,6 @@ abstract class Style<S extends Spec<S>> extends Mix<StyleSpec<S>>
     return provider?.style;
   }
 
-  /// Merges [other] into [current] while treating [IdentityStyle] as a no-op.
-  ///
-  /// This preserves the normal [merge] precedence where [other] wins, without
-  /// passing an [IdentityStyle] into generated styler merge methods.
-  @internal
-  static Style<S> mergeStyles<S extends Spec<S>>(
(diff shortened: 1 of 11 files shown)


tilucasoli:
LGTM

tilucasoli (approved):

PR #930, opened 2026-06-11
fix(mix): fall back to previous animation config on null transition

tilucasoli:
## Summary
- `StyleAnimationBuilder.didUpdateWidget` now passes `config ?? oldConfig` when (re)creating the driver, so a spec transition from an animated config to `animation: null` keeps animating with the previous config instead of jumping straight to the new value via a `NoAnimationDriver`.
- Added a widget test that captures the resolved color on every rebuild and asserts the value is neither the start nor the end mid-flight when switching from a `CurveAnimationConfig(red)` to `animation: null` (blue), then lands on blue once settled.

## Test plan
- [x] `fvm flutter test test/src/animation/style_animation_builder_test.dart` — all 10 tests pass.
- [x] Reverted the fix locally and reran the new test — it fails, confirming it actually exercises the regression.

### DIFF
--- a/packages/mix/lib/src/animation/style_animation_builder.dart
+++ b/packages/mix/lib/src/animation/style_animation_builder.dart
@@ -96,7 +96,7 @@ class _StyleAnimationBuilderState<S extends Spec<S>>
     } else {
       animationDriver.dispose();
       animationDriver = _createAnimationDriver(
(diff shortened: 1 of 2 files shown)


leoafarias (approved):
LGTM

PR #869, opened 2026-02-26
docs(website): Clean up and modernize all documentation examples

leoafarias:
## Summary

This PR removes internal `Prop` implementation details from user-facing documentation and standardizes all examples to use the fluent API. After comprehensive review of all 26 documentation pages, this update ensures users learn the correct public API without being exposed to internal plumbing.

## Changes

**Key Changes:**
- Removed internal `Prop` patterns from user-facing examples (introduction.mdx, design-token.mdx)
- Fluent API is now the primary pattern across all docs (BoxStyler().color(...).padding(...))
- Token call syntax used instead of Prop.token() ($primary() instead of Prop.token($primary))
- Prop.value() / Prop.token() shown only in number directives context (.multiply(), .clamp())
- Fixed animation API usage (springDurationBased has bounce parameter, not spring)
- Consistent use of Dart 3.11+ dot-shorthand in typed contexts
- Removed verbose class name prefixes (WidgetModifierConfig, AnimationConfig, BorderRadiusGeometryMix)

**Files Modified:** 20 documentation pages
- Core guides: introduction.mdx, design-token.mdx, directives.mdx, animations.mdx, widget-modifiers.mdx
- Tutorials: creating-a-widget.mdx, creating-context-variants.mdx, controlling-widget-state.mdx, theming.mdx
- Widget guides: box.mdx, flexbox.mdx, stack.mdx, icon.mdx, image.mdx, text.mdx, stylewidgets.mdx
- Overview pages: best-practices.mdx, getting-started.mdx, migration.mdx, utility-first.mdx

## Test Plan

- [x] Verified all documentation examples follow the fluent API pattern
- [x] Confirmed no internal Prop patterns remain in user-facing docs
- [x] Tested dot-shorthand patterns resolve correctly in typed contexts
- [x] Cross-referenced against Mix 2.0 API and styler-api-policy.md

## Verification Checklist

- [x] **Documentation Updates**: All examples use the fluent API with dot-shorthand where applicable
- [x] **Consistency**: Reviewed all 26 documentation pages for consistency
- [x] **API Correctness**: Verified animation methods and their parameters match actual API
- [x] **No Breaking Changes**: Documentation-only changes, no impact to code

PR #937, opened 2026-06-16
refactor(mix): convert AnimationConfig statics to factories and add spring wrappers

tilucasoli:
## Summary

Extracted from the still-open [PR #869](https://github.com/btwld/mix/pull/869). That PR was a bundle of documentation cleanup + this code change; the docs portion no longer applies here because the `website/` and `examples/` directories were moved out of this repo into [btwld/mix-docs](https://github.com/btwld/mix-docs). The animation_config change is the only piece that still belongs in mix, so it's lifted into its own focused PR.

- Converts `AnimationConfig.withDefaults`, `.springDescription`, `.spring`, and `.springWithDampingRatio` from `static` methods (using the unusual `.new(...)` dot-shorthand against `SpringAnimationConfig`) to proper `factory` constructors. This makes them consistent with the existing curve-based `factory AnimationConfig.X(...)` constructors above them in the same file.
- Adds seven new convenience factory wrappers on `AnimationConfig` so callers don't have to reach into the concrete subtypes:
  - `springDurationBased`, `curveSpring`, `curveSpringWithDampingRatio` (delegating to `CurveAnimationConfig`)
  - `springStandard`, `springWithDurationAndBounce`, `springCriticallyDamped`, `springUnderdamped` (delegating to `SpringAnimationConfig`)
- Tests:
  - Updates the existing `withDefaults` and `springDescription` tests with `as CurveAnimationConfig` / `as SpringAnimationConfig` casts (factories now return the sealed `AnimationConfig` base type).
  - Adds a new `AnimationConfig factory wrappers` group with 13 tests covering each new factory and parameter propagation.

## Related

The corresponding documentation cleanup from the source PR was ported to mix-docs in [btwld/mix-docs#15](https://github.com/btwld/mix-docs/pull/15).

## Test plan

- [x] `flutter analyze lib/src/animation/ test/src/animation/` — clean.
- [x] `flutter test test/src/animation/animation_config_test.dart` — 46/46 pass (33 existing + 13 new).
- [ ] Reviewer: run `melos run ci` / `melos run analyze` end-to-end if you want full-tree confirmation.

### DIFF
--- a/packages/mix/lib/src/animation/animation_config.dart
+++ b/packages/mix/lib/src/animation/animation_config.dart
@@ -309,21 +309,18 @@ sealed class AnimationConfig {
   const AnimationConfig();
 
   /// Creates animation data with default settings.
-  static CurveAnimationConfig withDefaults() {
-    return const CurveAnimationConfig(
-      duration: kDefaultAnimationDuration,
-      curve: Curves.linear,
-    );
-  }
+  factory AnimationConfig.withDefaults() => const CurveAnimationConfig(
+    duration: kDefaultAnimationDuration,
+    curve: Curves.linear,
+  );
 
   /// Creates a spring animation configuration with standard spring physics.
-  static SpringAnimationConfig springDescription({
+  factory AnimationConfig.springDescription({
     double mass = 1.0,
     double stiffness = 180.0,
     double damping = 12.0,
-    // Duration delay = Duration.zero,
     VoidCallback? onEnd,
-  }) => .new(
+  }) => SpringAnimationConfig(
     spring: SpringDescription(
       mass: mass,
       stiffness: stiffness,
@@ -332,12 +329,11 @@ sealed class AnimationConfig {
     onEnd: onEnd,
   );
 
-  static SpringAnimationConfig spring(
+  factory AnimationConfig.spring(
     Duration duration, {
     double bounce = 0,
-    // Duration delay = Duration.zero,
     VoidCallback? onEnd,
-  }) => .new(
+  }) => SpringAnimationConfig(
     spring: SpringDescription.withDurationAndBounce(
       duration: duration,
       bounce: bounce,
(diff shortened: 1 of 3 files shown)


packages/mix/test/src/animation/animation_config_test.dart
```diff
@@ -343,6 +344,123 @@ void main() {
         expect(config.spring.mass, 1.0);
         expect(config.spring.stiffness, 180.0);
       });
+
+      group('AnimationConfig factory wrappers', () {
```
leoafarias, line 347: **Test quality (non-blocking).** These wrappers are pure delegations, so the one regression worth guarding against is a transposed/renamed named arg — but most tests in this group assert only `isA<…>`: `springStandard`, `springCriticallyDamped`, `springUnderdamped`, and `curveSpringWithDampingRatio` check no propagated value, and `curveSpring`/`springDurationBased` check only `duration`. A swapped arg would keep them all green. Suggest asserting the forwarded values (the way the `springWithDampingRatio` test at L427-438 already does) and adding one `onEnd`-forwarding check — that path is currently untested for every new factory.

packages/mix/lib/src/animation/animation_config.dart
```diff
@@ -332,33 +329,126 @@ sealed class AnimationConfig {
...
+    duration,
+    mass: mass,
+    stiffness: stiffness,
+    ratio: ratio,
+    delay: delay,
+    onEnd: onEnd,
+  );
+
+  /// Creates a spring animation with standard physics parameters.
+  factory AnimationConfig.springStandard({
```
leoafarias: **API redundancy, worth settling pre-1.0 (non-blocking).** `springStandard` is functionally identical to `springDescription` (L318) — both build a `SpringAnimationConfig` from mass/stiffness/damping with the same defaults — and `springWithDurationAndBounce` (L419) is likewise identical to `spring` (L332). That's two public entry points for each result. Each wrapper also re-declares the default magic numbers (1.0 / 180.0 / 12.0) that already live on the subtype constructor, so the two copies can silently drift. Minor naming note: the prefix doesn't reveal the return type — `spring*` here returns `SpringAnimationConfig` (a real physics simulation) while `curveSpring*` / `springDurationBased` return `CurveAnimationConfig` (a `SpringCurve` sampled over a fixed duration), which is different runtime behavior. Since this is rc.0, it's the cheap moment to dedup/rename before the API freezes.

packages/mix/lib/src/animation/animation_config.dart
```diff
@@ -332,33 +329,126 @@ sealed class AnimationConfig {
...
     spring: SpringDescription.withDampingRatio(
       mass: mass,
       stiffness: stiffness,
       ratio: dampingRatio,
     ),
     onEnd: onEnd,
   );
+
+  /// Creates a curve-based spring animation with duration and bounce.
+  factory AnimationConfig.springDurationBased(
```
leoafarias, line 359: **Pre-existing, surfaced by these new wrappers (optional follow-up).** The curve-spring factories — this one plus `curveSpring` and `curveSpringWithDampingRatio` — return `CurveAnimationConfig`, whose `props` is `[duration, curve, delay]`, but `SpringCurve` (spring_curves.dart) doesn't implement value equality. So `springDurationBased(d) == springDurationBased(d)` is `false` for identical args (the `curve` field compares by identity). The existing equality test only covers `Curves.linear` (a const singleton), so it never surfaced. Adding these public wrappers widens the reach of the gap — might be worth giving `SpringCurve` value equality, or documenting the caveat.

leoafarias (commented):
A few inline notes on the static→factory conversion and the new wrappers — all non-blocking. Net: clean, low-risk, idiomatic refactor, and I verified every wrapper forwards its parameters to the correct subtype constructor (no mis-wires). Repo sweep shows no non-test consumers, so the conversion is internally safe.

PR #938, opened 2026-06-16
feat(mix): add ContextToken for zero-config context-derived token resolution

leoafarias:
## Summary

Introduces `ContextToken<T>`, a new token type that derives its value directly from `BuildContext` at resolution time — **no `MixScope` required**.

Previously, wiring a Material/Flutter theme value (e.g. `Theme.of(context).colorScheme.primary`) into a Mix style required declaring a token *and* providing it inside a `MixScope`. With `ContextToken` you define the resolver once and use it like any other token:

```dart
final primary = ContextToken<Color>(
  (context) => Theme.of(context).colorScheme.primary,
);

// No MixScope needed — just use it in a styler
final style = BoxStyler().color(primary());
```

`MixScope` still overrides the resolver when a matching entry exists, so the token remains fully compatible with the existing theming system.

## What changed

- **`ContextToken<T>`** — new `MixToken` subclass. `resolve()` calls `resolver(context)` directly unless a `MixScope` entry overrides it.
- **`MixScope.getToken`** — now accepts `T Function(BuildContext)` resolver entries in addition to plain `T` values, enabling context-derived overrides in scope.
- **`DoubleRef` sentinel registry** — hardened to monotonic sentinels with stable reverse-lookup, eliminating the hash-collision risk in the previous implementation. `DoubleRef._(double)` constructor is now private; the only public entry point is `DoubleRef.token(token)`.
- **`getReferenceValue`** — migrated from if-else chain to exhaustive `switch` expression; unsupported types now throw `UnsupportedError` with an actionable message instead of silently returning a `Prop`.
- **`BreakpointToken.resolve`** — replaced try/catch default-fallback with an explicit scope-presence check, surfacing misconfiguration instead of swallowing it.
- **`isAnyTokenRef`** — signature widened to `Object?`; null short-circuits cleanly.
- New integration test suite (`context_token_integration_test.dart`) covering resolver resolution, `MixScope` override, fluent-chain ordering, equality semantics, and all supported token types.

## Before / After

**Before** — required a scope declaration:
```dart
final primary = ColorToken('primary');

MixScope(
  tokens: {primary: Theme.of(context).colorScheme.primary},
  child: Box(style: BoxStyler().color(primary())),
)
```

**After** — resolver is the default, scope is optional:
```dart
final primary = ContextToken<Color>(
  (context) => Theme.of(context).colorScheme.primary,
);

Box(style: BoxStyler().color(primary()))
```

## Test plan

- [x] `melos run test:flutter` — **all 2747 tests passed** (run as `flutter test` in `packages/mix`)
- [x] New `context_token_integration_test.dart` (278 lines) covers happy path, chain ordering, `MixScope` override, equality, and multi-type resolution — **15 tests passed**
- [x] `melos run analyze` — **no issues** (`dart analyze`: _No issues found_; `dcm analyze --fatal-style --fatal-warnings`: _no issues found_)


packages/mix/lib/src/theme/tokens/token_refs.dart
```diff
@@ -237,52 +268,49 @@ final class BoxShadowListRef extends Prop<List<BoxShadow>>
...
+    _ => throw UnsupportedError(
+      'No token reference is registered for MixToken<$T> "${token.name}". '
+      'Either pick one of the supported token value types (Color, double, '
+      'Radius, Shadow, BoxShadow, TextStyle, Breakpoint, BorderSide, '
+      'FontWeight, Duration, List<Shadow>, List<BoxShadow>) or override '
+      'MixToken<$T>.call() in your subclass to return a custom reference.',
+    ),
+  };
+
+  return ref as T;
```
tilucasoli, line 315: This looks like a code smell. It's only used in the MixToken class — an abstract class used by other tokens. Would it be better to remove this function and instead implement it in each subtype as a required method?

PR #939, opened 2026-06-18
feat(mix_lint): add mix_avoid_token_ref_outside_mix rule

tilucasoli:
## What

Adds a new lint rule **`mix_avoid_token_ref_outside_mix`** to `mix_lint`.

Calling a `MixToken` — `token()` (or `token.mix()` on tokens that expose it) — does **not** return a concrete value. It returns a *reference* sentinel (`ColorRef`, `RadiusRef`, …) that only resolves later inside Mix's pipeline, against the surrounding `MixScope`. Passing that sentinel to a non-Mix API (a plain Flutter widget, a `dart:core` function) means it never resolves — a silent bug.

This rule flags exactly that escape.

## Behavior

- **Syntactic** trigger on the three ref producers: `token()`, `token.call()`, `token.mix()`. `token.resolve(context)` is exempt (it yields the real value).
- Fires **only in argument positions** — directly, or nested through collection literals. Assignment / return / bare statements are not flagged.
- **Allowed** when the innermost consuming invocation's receiver/target/constructed type descends from `Mix` (covers all Stylers and `*Mix` value utilities, whether shipped in `package:mix` or generated in a user package via `@MixableType`/`@MixableSpec`), with a `package:mix/` library-URI fallback for free functions.
- **Skipped** when the consumer type can't be resolved (`dynamic` / unresolved code).
- **Severity: `WARNING`** (first rule in the package above `INFO`) — a misrouted reference is a correctness bug. **No quick-fix**; the correction message points to `token.resolve(context)`.

### Don't
```dart
final primary = ColorToken('primary');
Container(color: primary()); // ref escapes, never resolves
```
### Do
```dart
final primary = ColorToken('primary');
Box(style: BoxStyler().color(primary()));        // pass to a Mix API
Container(color: primary.resolve(context));       // or resolve for non-Mix APIs
```

## Known gap

The check is syntactic: a reference first stored in a variable (`final c = token(); Container(color: c);`) is not detected. Documented in the README.

## Changes

- `lib/src/rules/mix_avoid_token_ref_outside_mix.dart` (new)
- `lib/src/utils/type_helpers.dart` — `isMixType()`
- `lib/main.dart` — registration
- `test/src/mix_stub.dart` — `Mixable`/`Mix` hierarchy, ref producers, non-Mix `Container`
- `test/src/rules/mix_avoid_token_ref_outside_mix_test.dart` (new) — 12 tests
- `README.md` — config entry + rule section

## Testing

- `dart test` — all 33 mix_lint tests pass (12 new).
- `dart analyze` — no issues.

### DIFF
--- a/packages/mix_lint/README.md
+++ b/packages/mix_lint/README.md
@@ -25,6 +25,7 @@ plugins:
     diagnostics:
       mix_avoid_defining_tokens_within_style: true
       mix_avoid_defining_tokens_within_scope: true
+      mix_avoid_token_ref_outside_mix: true
       mix_avoid_empty_variants: true
       mix_max_number_of_attributes_per_style: true
       mix_variants_last: true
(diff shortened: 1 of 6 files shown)


tilucasoli:
Added an example project: **`packages/mix_lint/example/`** — a minimal Flutter package that enables the plugin via `analysis_options.yaml` and demonstrates the rule in [`lib/token_ref_example.dart`](https://github.com/btwld/mix/blob/feat/mix-lint-token-ref-outside-mix/packages/mix_lint/example/lib/token_ref_example.dart):

- `BoxStyler().color(brandPrimary())` → ✅ OK (ref goes to a Mix API)
- `Container(color: brandPrimary.resolve(context))` → ✅ OK (resolved first)
- `Container(color: brandPrimary())` → ❌ **Warning** (ref escapes into a non-Mix API)

Also added `doc/mix_avoid_token_ref_outside_mix.md` for consistency with the other rules' docs.

The example resolves with `flutter pub get` and compiles cleanly; the rule's warning surfaces in the IDE/analysis server (not the `dart analyze` CLI). It's intentionally left out of the melos package globs, since `mix_lint` is deliberately excluded from the shared analyzer resolution — the example stands alone.

leoafarias (approved):
LGTM

PR #976, opened 2026-07-13
refactor(mix): reduce style pipeline overhead

leoafarias:
## Summary

- avoid subscribing a `StyleBuilder` to every widget-state aspect when it only needs to know whether a `WidgetStateProvider` exists
- bypass the no-op animation driver and `AnimatedBuilder` subtree when a style has no animation
- return immediately when a style declares no variants
- allocate pointer tracking and controller state lazily, and only for automatically produced hover/pressed states
- preserve controller state across external-to-internal transitions and cover the optimized paths with package-level regression tests

## Why

The rendering-pipeline investigation found several common paths doing work that could not affect output:

- the provider-presence check used a listening inherited-model lookup, so unrelated state changes could rebuild styles
- every `StyleBuilder` allocated a controller even when no interaction tracking was needed
- non-pointer widget-state variants installed a pointer detector/provider subtree that could never produce those states
- null-animation styles still created a no-animation driver and an `AnimatedBuilder`
- styles with no variants still entered filtering/list setup

The accepted changes remove that work without changing Mix's public styling API or legitimate animation and controller behavior. A package regression confirms that a hover update does not rebuild a pressed-only style, while a pressed update still rebuilds once.

The earlier combined workload measurement moved static Mix iterations by roughly 20–22 µs (about 4%, directional). That timing predates the final fresh-process protocol, so this PR treats it as characterization rather than a regression threshold; the stronger claim is the test-backed elimination of unnecessary dependencies, wrappers, controllers, and list work.

## What we discovered next

The investigation also identified larger headroom in variant processing:

- active-variant evaluation/merge is roughly 63–64% of realistic `Style.build` CPU
- generated property/spec resolution is roughly 30–33%
- two synchronous S2 state updates produce two controller notifications but coalesce into one Mix resolution/build; there is no duplicate pipeline computation

A fused variant-merge experiment made S0 21.5% faster and S0I 15.6% faster, but regressed S2 state transitions by 7.9% in 10/10 release pairs and also moved real-cadence profile results backward. A narrower fast path still regressed S2. Those experiments were reverted and are intentionally not part of this PR.

This leaves a clear follow-up target: optimize static variant evaluation or generated resolution without changing the already-efficient state-transition path.

## Validation

- `fvm dart run melos run gen:build`
- `fvm dart run melos run test:flutter --no-select` — 2,740 tests
- `fvm dart run melos run test:dart --no-select` — 320 generator tests and 37 linter tests
- `fvm dart run melos run analyze` — Dart and DCM analysis clean after bootstrapping the fresh worktree
- focused `StyleBuilder` and `StyleAnimationBuilder` suite — 33 tests
- `git diff --check`

## Scope

This is intentionally a code-only PR based directly on `main`: four Mix production files and one package test file. The separate benchmark harness and investigation documents are not included.

### DIFF
--- a/packages/mix/lib/src/animation/style_animation_builder.dart
+++ b/packages/mix/lib/src/animation/style_animation_builder.dart
@@ -32,17 +32,20 @@ class StyleAnimationBuilder<S extends Spec<S>> extends StatefulWidget {
 class _StyleAnimationBuilderState<S extends Spec<S>>
     extends State<StyleAnimationBuilder<S>>
     with TickerProviderStateMixin {
-  late StyleAnimationDriver<S> animationDriver;
+  StyleAnimationDriver<S>? _animationDriver;
 
   @override
   void initState() {
     super.initState();
     final spec = widget.spec;
     final config = spec.animation;
-    animationDriver = _createAnimationDriver(config: config, initialSpec: spec);
+    _animationDriver = _createAnimationDriver(
+      config: config,
+      initialSpec: spec,
+    );
   }
 
-  StyleAnimationDriver<S> _createAnimationDriver({
+  StyleAnimationDriver<S>? _createAnimationDriver({
     required AnimationConfig? config,
     required StyleSpec<S> initialSpec,
   }) {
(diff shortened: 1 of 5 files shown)


leoafarias:
Closing as superseded. The widget-state dependency check and the variant short-circuit from this branch landed on `main` in a more complete form through #1006 and #1016, which is why it now conflicts. The remaining idea, allocating the widget-state controller and animation driver lazily, would need a fresh implementation against the current `style_builder.dart` rather than a rebase. The `refactor/mix-style-pipeline-overhead` branch is kept, so #977, which is stacked on it, stays open.

PR #1005, opened 2026-08-04
feat(mix): rework Pressable input and semantics, add focus-visible variant

leoafarias:
## Summary

Interaction-subsystem changes to core `mix` extracted from the Tailwind parity branch (`leoafarias/mix-tailwind-compat`). They were *found* during parity work, but none of them are Tailwind-specific — they are standalone framework fixes plus two general-purpose features, so they ship as their own PR. The parity branch stacks on top of this one.

## Bug fixes (real bugs, reproducible without mix_tailwinds)

- **Nested widget-state discovery** (`style.dart`): `Style.widgetStates` only scanned top-level `WidgetStateVariant`s. A style whose only widget-state variants were nested (`onDark(BoxStyler().onHovered(...))`, `onBreakpoint(...)`) or negated (`onNot(hover)`) reported no dependencies, so `StyleBuilder` never attached a `MixInteractionDetector` and the variant **never activated**. Now recursive with identity-based cycle protection, and driven by the new `ContextVariant.widgetStateDependencies` so non-`WidgetStateVariant` variants (Not, FocusVisible) participate.
- **Duplicate pressed-state owners** (`pressable_widget.dart`): both `GestureDetector` (tapDown/tapUp/tapCancel) and `MixInteractionDetector`'s `Listener` wrote `pressed` to the same controller, with conflicting timing (e.g. tapCancel during a long-press cleared pressed while the pointer was still down). The Listener is now the single owner.
- **Controller lifecycle**: `late final _controller` ignored `widget.controller` swaps, and `dispose()` consulted the *current* widget to decide ownership — swapping external↔internal could dispose a controller the widget didn't own or leak the one it did. Fixed via `_ownsController` + `didUpdateWidget`.
- **Semantics correctness**: the semantics node exposed `onTap` even when disabled, had no enabled state or longPress action, and `GestureDetector`'s implicit semantics duplicated actions. Actions are now gated on `enabled`, `enabled` is exposed, and the GestureDetector is always excluded from semantics.
- **`PressableBox.enableFeedback`** existed but was never forwarded to `Pressable`; the full surface (cursor, focus, keyboard, controller, actions, semantics) is now forwarded.
- **Keyboard activation** previously bound `ActivateIntent` → `onPress` with no enabled check and no pressed feedback. Now Space/Enter/numpad Enter model held state: pressed on key down, activate **once** on key up, repeats suppressed, and cancellation on focus loss, disable, controller swap, and dispose.

## New features

- **`FocusVisibleVariant`** / `ContextVariant.focusVisible()` / `onFocusVisible(...)`: CSS `:focus-visible` equivalent, keyed off `FocusManager.highlightMode` (the same input-modality signal `FocusableActionDetector` uses), made reactive inside `Pressable` via the new internal `FocusHighlightModeProvider`.
- **`PressableSemanticsRole`** (button / link / none) + `semanticsLabel`.

`mix_protocol`'s schema inventory registers the new public symbols as v1 out-of-scope.

## Breaking changes

- `semanticButtonLabel` → `semanticsLabel` (no remaining usages in-repo).
- Deprecated `onKey` removed; use `onKeyEvent` (it runs first and can override activation).
- `Pressable` reserves Space/Enter/numpad Enter: custom `actions` no longer receive `ActivateIntent` for those keys. Deliberate contract so held-key state stays consistent; documented in the CHANGELOG.

Note one judgment call: activation fires on key-up for both Space and Enter (the web fires Enter on key-down). The uniform key-up model was chosen so the held/pressed lifecycle is consistent across keys.

## Test plan

- `packages/mix`: full suite passes (2,852 tests), including new coverage: keyboard hold/activate-once/cancel paths, controller-swap state transfer, `onKeyEvent` precedence, reserved-keys-vs-custom-actions, focus-visible modality switching, semantics role/action matrix, and nested/negated/cyclic variant discovery.
- `packages/mix_protocol`: inventory + contract tests pass.
- `dart analyze packages/mix packages/mix_protocol`: no issues.

### DIFF
--- a/packages/mix/CHANGELOG.md
+++ b/packages/mix/CHANGELOG.md
@@ -1,3 +1,33 @@
+## Unreleased
+
+### New features
+
+- **Typed focus-visible variants:** Added `FocusVisibleVariant`,
+  `ContextVariant.focusVisible()`, and `onFocusVisible(...)`. Context variants
+  can now declare their required states through
+  `ContextVariant.widgetStateDependencies`.
+- **Pressable semantics roles:** Added `PressableSemanticsRole` with button,
+  link, and neutral roles. `PressableBox` now forwards the full Pressable
+  focus, keyboard, controller, feedback, cursor, action, and semantics surface.
+
+### Breaking changes
+
+- **Pressable input and semantics:** Replaced `semanticButtonLabel` with
+  `semanticsLabel`, added `semanticsRole`, and removed the deprecated `onKey`
+  callback. Use `onKeyEvent` for custom keyboard handling.
+- **Reserved activation keys:** Pressable owns Space, Enter, and numpad Enter
+  so it can model held-key state consistently. Custom activation behavior must
+  use `onKeyEvent`; custom `actions` remain supported for other intents.
+
+### Fixes
+
+- **Nested widget-state discovery:** `StyleBuilder` now discovers state
+  requirements recursively through nested and negated variants with
+  identity-based cycle protection.
+- **Pressable lifecycle:** Pointer state has one owner, keyboard activation
+  fires once on key-up, cancellation clears held state, focus-visible follows
+  Flutter input modality, and disabled semantics expose no actions.
+
 ## 2.2.0-beta.1
 
 ### New features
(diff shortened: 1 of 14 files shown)


leoafarias:
Superseded by the split into #1006 (nested widget-state discovery fix, independently landable) and #1007 (Pressable input/semantics rework + focus-visible, stacked on #1006). Same combined tree, reorganized for review.

Issue #967, opened 2026-07-11
Fix style resolution correctness and deterministic variant ordering


leoafarias:
- Branch: `fix/style-resolution-correctness`
- Base: `origin/main`
- Package: `packages/mix`
- Priority: P0
- Change type: Correctness and hot-path cleanup

## Summary

Make inherited styles reactive, prevent unrelated widget-state changes from
re-resolving state-independent styles, remove dormant state-management work,
and make variant precedence deterministic without relying on an unstable sort.

This should preserve the current Style → StyleSpec → Widget architecture. It
is a focused correction to dependency ownership and ordering, not a pipeline
redesign.

## Why this matters

The current implementation has inverse dependency problems:

1. `Style.maybeOf` uses `getInheritedWidgetOfExactType`, so a `StyleBuilder`
   reads an inherited style without subscribing to `StyleProvider` changes. An
   otherwise stable child can keep stale resolved styling after its provider
   changes.
2. Every `StyleBuilder` calls `WidgetStateProvider.of(context)` with no aspect,
   even when its style has no widget-state variants. Hover, press, or other
   state changes can therefore re-resolve unrelated descendant styles.
3. Every `StyleBuilder` eagerly creates a `WidgetStatesController`, including
   styles that never install automatic interaction tracking.
4. `mergeActiveVariants` calls `List.sort` with only two priority values. Dart
   does not guarantee stable ordering for items that compare equal, even though
   Mix contract tests expect stored order to decide the winner within each
   priority bucket.
5. `WidgetState.scrolledUnder` is accepted by `ContextVariant.widgetState` and
   the controller extension, but `WidgetStateProvider.hasStateOf` always returns
   false for it.

## Evidence and source anchors

- `packages/mix/lib/src/core/style.dart`
  - `Style.maybeOf`
  - `Style.widgetStates`
  - `Style.mergeActiveVariants`
- `packages/mix/lib/src/core/style_builder.dart`
  - `_initController`
  - `_buildStyle`
  - `_StyleBuilderState.build`
- `packages/mix/lib/src/core/providers/style_provider.dart`
- `packages/mix/lib/src/core/providers/widget_state_provider.dart`
- `packages/mix/test/src/core/style_builder_test.dart`
- `packages/mix/test/src/core/style_get_all_variants_test.dart`
- `packages/mix/test/src/specs/pressable/widget_state_controller_test.dart`

A focused probe against the current checkout confirmed both runtime effects:
[...]

PR #1006, opened 2026-08-04
fix(mix): discover widget-state dependencies through nested variants

leoafarias:
## Summary

Fixes a silent activation bug in widget-state tracking, extracted from the Tailwind parity work (where `dark:hover:` combinations exposed it) but fully general.

## The bug

`Style.widgetStates` only scanned **top-level** `WidgetStateVariant`s. `StyleBuilder` uses that set to decide whether to attach a `MixInteractionDetector`, so a style whose only widget-state variants were:

- nested under another variant — `onDark(BoxStyler().onHovered(...))`, `onBreakpoint(...)`, or the reverse nesting, or
- negated — `onNot(ContextVariant.widgetState(.hovered), ...)`

reported **no** state dependencies, got no interaction detector, and those variants **never activated**. No error, no warning — the hover/press/focus style simply did nothing.

## The fix

- Collect dependencies recursively through nested **context-variant** branches, with identity-based cycle protection. Branches under un-applied named variants and `ContextVariantBuilder` placeholders are deliberately skipped: they cannot activate through `StyleBuilder` (which always resolves with empty `namedVariants`; `applyVariants` hoists applied branches to the top level), so walking them would mount an interaction detector for states that can never fire.
- Introduce `ContextVariant.widgetStateDependencies` (defaults to empty) so any context variant can declare the widget states it needs tracked: `WidgetStateVariant` reports its state, `NotVariant` delegates to its inner variant. This also gives future variants (e.g. a focus-visible variant, coming in the stacked follow-up PR) a first-class way to participate in tracking.
- Mount `MixInteractionDetector` only for states it can actually drive from pointer input (`hovered`/`pressed`, declared as `MixInteractionDetector.pointerDrivenStates`). States such as `disabled` and `focused` can only come from an external `WidgetStatesController` or an ancestor scope, so styles depending solely on those no longer gain an opaque hit-test target that swallowed pointer events aimed at widgets beneath them, and no longer hijack the state scope of descendants that do track hover. The ancestor-scope existence check now uses `getInheritedWidgetOfExactType` so it no longer subscribes `StyleBuilder` to every state change.

## Issue-tracker context

No existing issue reports this. #967 lists `Style.widgetStates` among its anchors but describes the opposite family of defects (over-subscription and stale inherited styles); this PR addresses the under-tracking side and leaves #967's concerns untouched.

## Test plan

- New widget tests: negated hover, hover nested under breakpoint, hover nested under dark mode, reversed nesting, nested pressed, and a cyclic-structure unit test.
- Detector-mounting tests: no-widget-state and `onEnabled`-only styles stay transparent to taps, nested hover mounts the detector, an ancestor with only non-pointer states no longer hijacks a descendant's hover scope, and the scope existence check does not subscribe to unrelated state changes.
- Regression tests for the inert named-variant shape: hover under an un-applied named variant reports no dependencies and stays transparent to taps; `applyVariants` hoisting is still discovered.
- Full `packages/mix` suite on this branch: **2,848 tests pass**, `dart analyze` and `dart format` clean.

Stacked follow-up: #1005's replacement feature PR (Pressable input/semantics rework + focus-visible) is based on this branch.

### DIFF
--- a/packages/mix/CHANGELOG.md
+++ b/packages/mix/CHANGELOG.md
@@ -11,10 +11,12 @@
 ### Fixes
 
 - **Nested widget-state discovery:** `Style.widgetStates` now discovers state
-  requirements recursively through nested and negated variants with
-  identity-based cycle protection, so variants like
+  requirements recursively through nested and negated context-variant branches
+  with identity-based cycle protection, so variants like
   `onDark(BoxStyler().onHovered(...))` and `onEnabled(...)` are tracked instead
-  of silently never activating.
+  of silently never activating. Branches under un-applied named variants are
+  deliberately not tracked because they cannot activate until `applyVariants`
+  hoists them to the top level.
 - **Interaction detector is mounted only when it can help:** `StyleBuilder` now
   installs its pointer-interaction detector only for the states that detector
   actually drives (`hovered`/`pressed`). States such as `disabled` and `focused`
--- a/packages/mix/CHANGELOG.md
+++ b/packages/mix/CHANGELOG.md
@@ -1,3 +1,11 @@
+## Unreleased
+
+### Fixes
+
+- **Nested widget-state discovery:** `StyleBuilder` now discovers state
+  requirements recursively through nested and negated variants with
+  identity-based cycle protection.
+
 ## 2.2.0-beta.1
 
 ### New features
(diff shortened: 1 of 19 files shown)


tilucasoli (approved):
LGTM!

Issue #314, opened 2024-06-21
`PressableBox` not pressed when using keyboard trigger `onPress`


definev:
### Package version

1.1.0

### Flutter version

Flutter 3.22.2 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 761747bfc5 (2 weeks ago) • 2024-06-05 22:15:13 +0200
Engine • revision edd8546116
Tools • Dart 3.4.3 • DevTools 2.34.3

### Steps to reproduce

I create two PressableBox. Use my keyboard to traversal and press enter to press it but the `onPress` not triggered

### Expected results

Press with enter key

### Actual results

Not call `onPress`

### Code sample

<details open><summary>Code sample</summary>

```dart
```

</details>

PR #1007, opened 2026-08-04
feat(mix): rework Pressable input and semantics, add focus-visible variant

leoafarias:
## Summary

Pressable interaction-subsystem rework extracted from the Tailwind parity branch (`leoafarias/mix-tailwind-compat`). Found during parity work, but none of it is Tailwind-specific: standalone input/lifecycle/a11y fixes plus two general features. Replaces #1005 together with its base PR.

**Stacked on** `fix/nested-widget-state-discovery`: `FocusVisibleVariant` declares its focus dependency through the `widgetStateDependencies` API introduced there. This PR's diff shows only its own changes; it retargets to `main` automatically when the base merges.

## Bug fixes

- **Duplicate pressed-state owners**: both `GestureDetector` (tapDown/tapUp/tapCancel) and `MixInteractionDetector`'s `Listener` wrote `pressed` to the same controller, with conflicting timing (e.g. tapCancel during a long-press cleared pressed while the pointer was still down). The Listener is now the single owner.
- **Keyboard activation** previously bound `ActivateIntent` → `onPress` with no enabled check and no pressed feedback. Now Space/Enter/numpad Enter model held state: pressed on key down, activate **once** on key up, repeats suppressed, cancellation on focus loss, disable, controller swap, and dispose. Refs #314 — the "onPress not firing" half was fixed back then; this completes the missing pressed-state half its title describes.
- **Controller lifecycle**: `late final _controller` ignored `widget.controller` swaps, and `dispose()` consulted the *current* widget to decide ownership — swapping external↔internal could dispose a controller the widget didn't own or leak the one it did. Fixed via `_ownsController` + `didUpdateWidget`.
- **Semantics correctness**: the semantics node exposed `onTap` even when disabled, had no enabled state or longPress action, and `GestureDetector`'s implicit semantics duplicated actions. Actions are now gated on `enabled`, `enabled` is exposed, and the GestureDetector is always excluded from semantics.
- **`PressableBox.enableFeedback`** existed but was never forwarded to `Pressable`; the full surface (cursor, focus, keyboard, controller, actions, semantics) is now forwarded.

## New features

- **`FocusVisibleVariant`** / `ContextVariant.focusVisible()` / `onFocusVisible(...)`: CSS `:focus-visible` equivalent, keyed off `FocusManager.highlightMode` (the same input-modality signal `FocusableActionDetector` uses), made reactive inside `Pressable` via the new internal `FocusHighlightModeProvider`.
- **`PressableSemanticsRole`** (button / link / none) + `semanticsLabel`.

`mix_protocol`'s schema inventory registers the new public symbols as v1 out-of-scope.

## Breaking changes

- `semanticButtonLabel` → `semanticsLabel` (no remaining usages in-repo).
- Deprecated `onKey` removed; use `onKeyEvent` (it runs first and can override activation).
- `Pressable` reserves Space/Enter/numpad Enter: custom `actions` no longer receive `ActivateIntent` for those keys. Deliberate contract so held-key state stays consistent; documented in the CHANGELOG.

Note one judgment call: activation fires on key-up for both Space and Enter (the web fires Enter on key-down). The uniform key-up model was chosen so the held/pressed lifecycle is consistent across keys.

## Issue-tracker context

Searched open and closed issues (Pressable, pressed, focus, keyboard, semantics, accessibility, enableFeedback, focus-visible, …): no open issue reports these bugs or requests these features. #314 (closed, v1-era) is the nearest prior report, referenced above.

## Test plan

- Combined tree (this branch): full `packages/mix` suite passes (**2,852 tests**), including new coverage: keyboard hold/activate-once/cancel paths, controller-swap state transfer, `onKeyEvent` precedence, reserved-keys-vs-custom-actions, focus-visible modality switching, and a semantics role/action matrix (enabled × tap × longPress).
- `packages/mix_protocol` inventory + contract tests pass; `dart analyze` clean.

The Tailwind parity branch consumes this PR (`focus-visible:` prefix → `onFocusVisible`, `TwPressable` → semantics roles) and will be rebased once it lands.

### DIFF
--- a/packages/mix/CHANGELOG.md
+++ b/packages/mix/CHANGELOG.md
@@ -16,20 +16,23 @@
   `semanticsLabel`, added `semanticsRole`, and removed the deprecated `onKey`
   callback. Use `onKeyEvent` for custom keyboard handling.
 - **Reserved activation keys:** While it holds primary focus and can activate,
-  Pressable owns the keys Flutter maps to `ActivateIntent` — Space, Enter,
-  numpad Enter, select, and game button A — so it can model held-key state
-  consistently. Custom activation behavior must use `onKeyEvent`; custom
-  `actions` remain supported for other intents. Those keys are left untouched
-  when a descendant holds focus, so nested text fields keep working.
+  Pressable owns unmodified Space, Enter, numpad Enter, select, and game button
+  A so it can model held-key state consistently. Override those direct key
+  bindings with `onKeyEvent`; custom `actions` remain available to other
+  shortcuts and programmatic intents. Those keys are left untouched when a
+  descendant holds focus, and modified chords are left to application
+  shortcuts.
 
 ### Fixes
 
 - **Nested widget-state discovery:** `StyleBuilder` now discovers state
   requirements recursively through nested and negated variants with
   identity-based cycle protection.
-- **Pressable lifecycle:** Pointer state has one owner, keyboard activation
-  fires once on key-up, cancellation clears held state, focus-visible follows
-  Flutter input modality, and disabled semantics expose no actions.
+- **Pressable lifecycle:** Pointer and keyboard press sources are combined
+  without clearing each other, keyboard activation fires once on key-up,
+  cancellation clears held state, focus-visible follows Flutter input modality,
+  and disabled controls ignore custom key handling and expose neither semantic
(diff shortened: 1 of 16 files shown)


tilucasoli (approved):
LGTM

PR #1025, opened 2026-08-13
fix(mix_generator): allow plain MixableSpec widget targets

### LINKED ISSUE #1022: @MixableSpec(target:) rejects plain Widget targets with compatible style parameters
### Package version

- `mix_generator 2.2.0-beta.2`
- `mix_annotations 2.2.0-beta.1`
- `mix 2.2.0-beta.4`

### Flutter version

Flutter 3.44.0, Dart 3.12.0.

### Steps to reproduce

1. Create a Flutter package with the versions above and `build_runner`.
2. Enable the spec Styler builder:

   ```yaml
   targets:
     $default:
       builders:
         mix_generator|spec_styler_generator:
           enabled: true
   ```

3. Save the code sample below as `lib/repro.dart`.
4. Run `dart run build_runner build`.

I ran this repro from a clean builder state. The command exited with code 1 and the diagnostic shown below.

### Expected results

`PlainWidget` is a `Widget` with a named `style` parameter that accepts `Style<PlainSpec>`, so it should be a valid `@MixableSpec(target:)`. The generated Styler only needs to construct `PlainWidget(..., style: this)`; that body does not depend on `StyleWidget`.

This shape is already supported by `@MixWidget(target:)`: [`_directTargetCallSource`](https://github.com/btwld/mix/blob/726d85e84a4188ccb547d09cbdb7020c729574bd/packages/mix_generator/lib/src/mix_widget_generator.dart#L316-L410) validates a plain `Widget` target and its `style` parameter, and [`MixWidgetBuilder._writeDirectTargetBuild`](https://github.com/btwld/mix/blob/726d85e84a4188ccb547d09cbdb7020c729574bd/packages/mix_generator/lib/src/core/builders/mix_widget_builder.dart#L143-L164) emits the direct constructor call. Both paths already use the shared constructor-parameter planning helpers; the `@MixableSpec` path then renders the equivalent body through [`extractCallParams` and `renderWidgetCall`](https://github.com/btwld/mix/blob/726d85e84a4188ccb547d09cbdb7020c729574bd/packages/mix_generator/lib/src/core/helpers/widget_call_planner.dart#L182-L247).

### Actual results

Generation fails before validating whether the named `style` parameter is compatible:

```text
Widget PlainWidget must extend StyleWidget<PlainSpec> to be used as @MixableSpec(target:).
```

The rejection comes from [`validateMixableSpecTargetConstructor`, lines 136–146](https://github.com/btwld/mix/blob/726d85e84a4188ccb547d09cbdb7020c729574bd/packages/mix_generator/lib/src/core/helpers/widget_call_planner.dart#L129-L146).

### Code sample

<details open><summary>Code sample</summary>

```dart
import 'package:flutter/foundation.dart';
import 'package:flutter/widgets.dart';
import 'package:mix_annotations/mix_annotations.dart';

part 'repro.g.dart';

@MixableSpec(target: PlainWidget.new)
final class PlainSpec with _$PlainSpec {
  const PlainSpec({this.opacity});
  final double? opacity;
}

class PlainWidget extends StatelessWidget {
  const PlainWidget({super.key, required this.style, required this.label});
  final Style<PlainSpec> style;
  final String label;

  @override
  Widget build(BuildContext context) => Text(label);
}
```

</details>

### Proposed change

Replace the `StyleWidget<Spec>` supertype requirement with validation that the target has a named `style` parameter whose type accepts the generated Styler, analogous to the existing `@MixWidget(target:)` validation. I would be happy to send a PR.

Found while working on a design system with 32 plain-widget targets of this shape.

leoafarias:
#### Related issue

Closes #1022.

### Description

`@MixableSpec(target:)` now accepts plain `Widget` constructors when their named `style` parameter can receive the generated Styler, removing the unnecessary `StyleWidget` inheritance requirement.

### Changes

- Validate the target's `Widget` subtype and constructor contract while preserving existing parameter and generic-target handling.
- Cover compatible `Style<Spec>`, `Object`, and exact generated-Styler parameters, including when a prior generated part is already resolved.
- Match `Style<Spec>` by analyzer element and library identity so unrelated same-named types are rejected.
- Qualify target-class static defaults when copied into generated `call()` methods, sharing the established `@MixWidget(target:)` behavior while leaving top-level defaults unchanged.
- Keep the exact-type shortcut limited to concrete generated Stylers so legacy `@MixableStyler` mixins retain their stricter `Style<Spec>` receiver contract.
- Update the changelog, generator documentation, and regression coverage for valid and invalid target shapes.

**Review Checklist**

- [x] **Testing**: Generator, repository, clean-build, prior-output, and generated-output resolver tests were run after merging the latest `main`.
- [x] **Breaking Changes**: No breaking changes; existing `StyleWidget` targets remain supported.
- [x] **Documentation Updates**: The generator README, changelog, and Mix skill documentation are updated.
- [x] **Website Updates**: No website update is required for this generator validation fix.

`melos run gen:build` and `melos run ci` passed, both protocol inventories passed, all 11 packages passed Dart analysis, and the changed planner passed focused DCM analysis. The aggregate analysis command remains non-zero only for 11 pre-existing DCM findings on unchanged lines.


leoafarias:
I found a resolution-state edge case that looks blocking for the exact generated-Styler parameter shape this PR adds.

On a clean build, a parameter such as `BoxStyler style` is unresolved and reaches `_targetStyleAcceptsGeneratedStyler` as `InvalidType`, so it is accepted. When a prior generated part is separately supplied to or resolved by the generator, `BoxStyler` instead becomes an `InterfaceType`. The helper then finds its `Style<BoxSpec>` supertype and checks whether that base type is assignable to `BoxStyler`; it is not, so generation fails with:

```
@MixableSpec(target:) PlainWidget `style` parameter cannot accept the generated `BoxStyler`.
```

I reproduced this by including a prior generated part containing `BoxStyler extends Style<BoxSpec>` and running the part builder to a separate output. The current clean-build test passes, while the prior-output case fails. I could not reproduce this through a normal incremental `build_runner` run, so this should not be read as a claim that every later build fails; the underlying problem is that validation changes with analyzer resolution state.

This is relevant to checked-in or otherwise independently resolved generated parts. Remix has that shape—its plain widgets type `style` as concrete generated classes such as `ButtonStyler`—so the promised behavior should remain valid whenever those types are already resolved.

A localized fix is to pass `stylerName` into the compatibility helper and, before the `Style<Spec>` supertype check, accept an `InterfaceType` whose element has that name and belongs to the spec's library. I prototyped that together with a prior-output regression test; the same-build case and the existing custom-style-subtype and wrong-spec negative tests still pass.

leoafarias:
Follow-up from validating the current head (`367b1f9`) against the full Remix workspace: the earlier analyzer-resolution-state issue is fixed, but the integration exposed one additional blocking target-default case.

When a target constructor uses an unqualified static member from its own class as a default, for example `transitionBuilder = defaultAccordionTransitionBuilder` on `RemixAccordion`, the generated Styler method copies that expression into the Styler class unchanged. The result references `defaultAccordionTransitionBuilder` out of scope and does not compile; it needs to be emitted as `RemixAccordion.defaultAccordionTransitionBuilder`.

The `@MixWidget` path already qualifies direct target-member defaults. Sharing that qualification with the `@MixableSpec(target:)` planner, plus a resolver regression fixture for prior generated output, fixes the case locally. With that patch, `mix_generator` analysis and all 372 tests pass. [Remix #152](https://github.com/conceptadev/remix/pull/152) validates all 27 migrated component surfaces and the full 3,028-test workspace; it temporarily normalizes Accordion's nullable default inside the constructor so it can remain pinned to this exact PR head.

Verdict remains: the overall direction is correct, but this static-default qualification should be fixed before #1025 merges so other same-library targets cannot generate uncompilable code.

